PR #154, opened 2022-06-19
Extract subscribe

victimsnino:


src/rpp/rpp/observables/specific_observable.hpp
```diff
@@ -49,52 +47,11 @@ class specific_observable : public interface_observable<Type, specific_observabl
...
-     * \details this overloading accepts raw functions to construct specific subscriber with specific observer
-     * \return subscription on this observable which can be used to unsubscribe
-     */
-    template<typename ...Args>
-        requires constraint::specific_subscriber_constructible<Type, std::decay_t<Args>...>
-    composite_subscription subscribe(Args&&...args) const
-    {
-        return subscribe_impl(rpp::make_specific_subscriber<Type>(std::forward<Args>(args)...));
-    }
+    friend struct details::member_overload<Type, specific_observable<Type, OnSubscribeFn>, details::subscribe_tag>;
```
tcw165, line 50: nit: I like the clean modularization but I do feel keeping `subscribe()` in place in the `*_observable` classes is helpful:
- `subscribe()` is unlike other operators that may be optional, it needs to be easily found in the class for readability. (actually same as `lift()`.
- Not sure if the inheritance would blow up the binary size. I guess each time when a class derives from a new class, 2~4 bytes for unique address pointer are introduced. And this is a template framework, the size could easily mount up. e.g. In some embedded system, 1K is even like luxury.

tcw165: Can we use tag dispatching or SFINAE trick to replace the inheritance? That could probably save more space?

victimsnino: The reason for extraction of subscribe is reusability:
1) Currently specific_observable just has most rich overloading for subscribe method
2) Also i have blocking observable which in most reuse same code with only different "subscribe_impl" logic, but it also need same amount of overloading
3) I've planned to start `group_by` operator which also need extra "observable" with extra subscribe method.

As a result, we have 3 classes with SAME overloadings and only different impl logic. So, i've extracted only overloading over input data and placed implementation in the final observable.

victimsnino: > Can we use tag dispatching or SFINAE trick to replace the inheritance? That could probably save more space?

Not sure if it really makes sense. We can return to the code as in rxcpp, but it really doesn't make sense:
1) each operator accepts `(args...)` and you MUST to KNOW arguments for this one instead of checking tip from ide
2) current inheritance mode doesn't use vtable at all, so, it is cheap enough

```cpp
struct NoInheritance{};

struct Base{};
struct Der : public Base{};

struct BaseVtable{virtual ~BaseVtable() = default;};
struct DerVtable : public BaseVtable{};


template<typename T>
struct interface_observable{};

struct specific_observable : interface_observable<specific_observable>{};


int main() {
    std::cout << "sizeof(NoInheritance) " << sizeof(NoInheritance) << std::endl;
    std::cout << "sizeof(Base) " << sizeof(Base) << std::endl;
    std::cout << "sizeof(Der) " << sizeof(Der) << std::endl;
    std::cout << "sizeof(DerVtable) " << sizeof(DerVtable) << std::endl;
    std::cout << "sizeof(specific_observable) " <<  sizeof(specific_observable) << std::endl;
}

//

sizeof(NoInheritance) 1
sizeof(Base) 1
sizeof(Der) 1
sizeof(DerVtable) 8
sizeof(specific_observable) 1
```

victimsnino: > Can we use tag dispatching or SFINAE trick to replace the inheritance? That could probably save more space?

I will think about it

victimsnino: > Can we use tag dispatching or SFINAE trick to replace the inheritance? That could probably save more space?

Currently i'm thinking only about macro/includes inside observable body....

```cpp
// specific_observable.hpp

class specific_observable
{
public:
....
 MAP_DECLARATION();
 FILTER_DECLARATION();
};


// map.hpp
#define MAP_DECLARATION() \
auto map(.....).......
```

victimsnino: Looks like gcc/clang optimizes it efficiently and after removing virtual destructior from virtual_observable size of observable is collapsed to sizeof OnSubscribeFn. Unfortunately, MSVC does it less effictive and adds +1 for each inheritance

victimsnino: Fixed in #158

PR #155, opened 2022-06-19
Add subscription check before activating the OnSubscribeFn of an observable

tcw165:
We’d like to not activate the observable when it’s subscribed to a dead subscriber for `flat_map()` or `switch_map()` operators.


victimsnino:
Are you going to submit it? Looks like currently your changes fail some tests

PR #157, opened 2022-06-20
Get rid of member_overload

victimsnino:


victimsnino:
@tcw165 , what do you think about this approach?

victimsnino:
Looks like way with #include brokes doxygen. Way with macros also is really non-readable. Probably best way is to move declarations back to the base class...

PR #170, opened 2022-07-14
Allow IDE like CLion/VSCode to parse C++20 code via an executable target

tcw165:


victimsnino:
Why can't you just set cmake variable locally when you generate solution?

tcw165:
Per offline discussion, we decided to add section in README to guide developers to add `RPP_BUILD_SAMPLES` flag to their IDE(s) so that the code parsing could work.

PR #173, opened 2022-07-15
Add auto-cleanup by default for `make_child`

victimsnino:


src/rpp/rpp/subscriptions/composite_subscription.hpp
```diff
@@ -45,6 +45,8 @@ class composite_subscription final : public subscription_base
         {
             if (const auto pstate = get_state())
                 static_cast<state*>(pstate)->add(sub);
+            else
+                sub.unsubscribe();
```
tcw165, line 49: Can you add a test case for this `else` case?

PR #177, opened 2022-07-15
Allocate one "never" observable per deduced type.

tcw165:


victimsnino:
Are you sure that it will give some boost? For me it is really trivial action to construct observable with empty on_subscribe

tcw165:
I reference to [RxJava implementation for never](https://github.com/ReactiveX/RxJava/blob/3.x/src/main/java/io/reactivex/rxjava3/core/Observable.java#L3900) (or [declaration link](https://github.com/ReactiveX/RxJava/blob/5bb119b1489afe4fbc5bb51b828be95971ee8f1d/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableNever.java)), where they declare that as an static instance.

But they use `Object` as type parameter in Java. 

> Are you sure that it will give some boost?

I think it would save the allocation for never-observable but hard to notice how much we could save honestly. Both never and empty are useful in difference cases.

For example on `never()`:

```
someUpstream
  .switch_,map([](const auto& val) {
    if (val == some_criteria) {
      return rpp::sources::never().as_dynamic();
    }

    return nextActionObservable();
  })
```

Here, using `never()` inside a `switch_map()` is an alternative way to using `.filter(...)`. However, using `never()` could reduce length of chain.

victimsnino:
I see your point, but it doesn’t work in such a way :-) Anyway allocation would happens because static object inside function must be copied to return it from function. So, instead of construction you would do copy :-). One way to to avoid it: return const& on static object. In this case yes, object would be created only once and no any copies… but even in this case I’m not sure if it is faster…

tcw165:
> Anyway allocation would happens because static object inside function must be copied to return it from function.

I see. Yup, the copy is happening.

> anyway, to see performance boost it is better to create benchmark ;-)

I'd like to learn to create a bench mark. Can you point me out some examples?

victimsnino:
All RPP benchmarks have paced there 

https://github.com/victimsnino/ReactivePlusPlus/blob/main/src/tests/benchmarks.cpp

victimsnino:
Any updates related to this one?

victimsnino:
Delayed till we really need it

PR #185, opened 2022-07-25
Add 'buffer' operator

tcw165:


src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,83 @@
...
+{
+
+/// A non-copyable class that provides a copyable on_next for the subscriber and
+/// allows copies of on_next(s) to share the same states.
+template<constraint::decayed_type UpstreamType, constraint::decayed_type DownstreamType>
+struct buffer_on_next : public std::enable_shared_from_this<buffer_on_next<UpstreamType, DownstreamType>>
+{
+    using DownstreamValueType = UpstreamType;
+
+    explicit buffer_on_next(const size_t& count) : max_(count) {};
```
victimsnino, line 31: It is better to avoid references for PODs

```suggestion
    explicit buffer_on_next(size_t count) : max_(count) {};
```

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,83 @@
...
+            {
+                auto copy = std::vector<DownstreamValueType>(shared_this->buckets_);
+                shared_this->buckets_.clear();
+                subscriber.on_next(std::move(copy));
+            }
+        };
+    }
+
+private:
+    const size_t max_;
```
victimsnino, line 59: You can avoid it in the next way:
1) During construction you need to `.reserve()` space in your vector (+ it will increase performance due to preallocated memory)
2) During `on_next` you can compare size with `.capacity()`

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,83 @@
...
+    buffer_on_next& operator=(const buffer_on_next&) = delete;
+    buffer_on_next& operator=(buffer_on_next&&) noexcept = default;
+
+    /// \return a on_next function for subscriber, that is copyable and all the
+    /// copies share the same state.
+    auto get_on_next() {
+        return [shared_this = this->shared_from_this()](
+                const UpstreamType& value,
+                const constraint::subscriber_of_type<DownstreamType> auto& subscriber)
+        {
```
victimsnino, line 47: Add check "for an idiot" over max_size == 0. In this case just do return to prevent infinite grow of vector

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,83 @@
...
+        auto state = std::make_shared<buffer_on_next<UpstreamType, DownstreamType>>(count);
+        auto subscription = subscriber.get_subscription().make_child();
+
+        return create_subscriber_with_state<UpstreamType>(
+            std::move(subscription),
+            std::forward<decltype(subscriber)>(subscriber),
+            // Get a copy of on_next that shares the same state.
+            state->get_on_next(),
+            forwarding_on_error{},
+            forwarding_on_completed{});
```
victimsnino, line 79: You need to handle `on_completed` to send all collected but not yet emitted items. For example, when your count == 5, but collected only 4 items and on_completed happens

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+template<constraint::decayed_type UpstreamType, constraint::decayed_type DownstreamType>
+struct buffer_impl
+{
+    const size_t count;
+
+    auto operator()(auto&& subscriber) const
+    {
+        auto state = std::make_shared<buffer_on_next<UpstreamType, DownstreamType>>(count);
+        auto subscription = subscriber.get_subscription().make_child();
```
victimsnino, line 71: You don't need `.make_child()` there. `.make_child()` used when upstream can live shorter than downstream. In case of same lifetime use original subscription

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,83 @@
...
+    const size_t max_;
+    std::vector<DownstreamValueType> buckets_;
+};
+
+template<constraint::decayed_type UpstreamType, constraint::decayed_type DownstreamType>
+struct buffer_impl
+{
+    const size_t count;
+
+    auto operator()(auto&& subscriber) const
```
victimsnino, line 68: Add check over subscriber to avoid some template issues in future
```suggestion
   
    template<constraint::subscriber_of_type<DownstreamType> TSub>
    auto operator()(TSub&& subscriber) const
```

src/samples/doxygen/buffer.cpp
```diff
@@ -0,0 +1,52 @@
+#include <rpp/rpp.hpp>
+#include <iostream>
+
+std::ostream & operator<<(std::ostream & out, const std::vector<int>& list)
+{
+    out << "{";
+    auto size = list.size();
+    for (int i = 0; i < size; ++i)
```
victimsnino, line 8: Not critical in this case, but just a tip: use size_t when you deal with size of containers

src/samples/doxygen/buffer.cpp
```diff
@@ -0,0 +1,52 @@
...
+        // The stream that uses lvalue overloads for operators
+        auto source = rpp::source::just(1, 2, 3, 4, 5);
+        source.buffer(2)
+            .subscribe(
+                [](const std::vector<int> &v) { std::cout << v << "-"; },
+                [](const std::exception_ptr &error) {},
+                []() { std::cout << "|" << std::endl; });
+        // Source: -1-2-3-4-5--|
+        // Output: {1,2}-{3,4}-|
+    }
```
victimsnino, line 49: Any difference?

src/tests/rxcpp_benchmark.cpp
```diff
@@ -310,6 +310,21 @@ TEST_CASE("merge")
...
 
+TEST_CASE("buffer")
+{
+    BENCHMARK_ADVANCED("buffer")(Catch::Benchmark::Chronometer meter)
+    {
+        auto sub = rxcpp::make_subscriber<std::vector<int>>();
+
+        meter.measure([&]
+            {
+                return rxcpp::observable<>::range(1, 5)
```
victimsnino, line 321: To have correct benchmark results use same code. rpp uses `just(...)`, so, use `just` in rxcpp too

src/tests/test_buffer.cpp
```diff
@@ -0,0 +1,156 @@
...
+                });
+                CHECK(mock.get_on_completed_count() == 0);
+                CHECK(mock.get_on_error_count() == 1);
+            }
+        }
+        WHEN("subscribe on it via buffer(2)")
+        {
+            obs.flat_map([](auto v) {
+                  return v;
+                })
```
victimsnino, line 109: I think, flat_map can be as a part of `obs`

victimsnino: Also you can change flat_map to merge =)

src/tests/test_buffer.cpp
```diff
@@ -0,0 +1,156 @@
...
+                .buffer(2)
+                .subscribe(mock);
+            THEN("shall see -{1}-x, which means error event is through")
```
victimsnino, line 112:
```suggestion
            THEN("shall see --x, which means error event is through")
```

victimsnino (changes requested):

src/rpp/rpp/operators/fwd/buffer.hpp
```diff
@@ -0,0 +1,80 @@
...
+    using DownstreamType = std::vector<Type>;
+
+    /**
+    * \brief periodically gather items emitted by an Observable into bundles and emit these bundles rather than emitting
+    * the items one at a time
+    *
+    * \marble buffer
+        {
+            source observable    : +-1-2-3-|
+            operator "buffer(2)" : +---{12}-|
```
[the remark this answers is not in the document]
tcw165: Oh, that's interesting. I double check the docs in reactivex and didn't find any description for such case. Do you know which one is the right expectation?

src/rpp/rpp/operators/buffer.hpp
tcw165: Good idea!

src/rpp/rpp/operators/buffer.hpp
tcw165: Thanks, I'll do that. :)

src/rpp/rpp/operators/buffer.hpp
tcw165: Sounds good. For my education, what are those potential issues?

src/rpp/rpp/operators/buffer.hpp
tcw165: Oh I see. I'll update the code.

src/rpp/rpp/operators/buffer.hpp
tcw165: Yup, you mention the behavioral expectation in other thread. Let's follow up that thread and if your expectation is the right answer, I'll update the code.

src/rpp/rpp/operators/fwd/buffer.hpp
```diff
@@ -0,0 +1,80 @@
...
+        }
+    *
+    * \details It uses lift overload for subscriber <=> subscriber
```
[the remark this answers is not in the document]
tcw165: Sounds good! :)

src/samples/doxygen/buffer.cpp
tcw165: Oh yep, I should have used `auto` for compiler to automatically match the type.

src/samples/doxygen/buffer.cpp
tcw165: The only difference is the 1st chain uses the rvalue overload and 2nd chain uses the lvalue. I just want to point out they are different overloads and people could break one while updating the other.

src/tests/rxcpp_benchmark.cpp
tcw165: That's fair. Besides, I thought I hit the error that there's no matching function for `just(1,2,3,4,5,6,7)`.

src/tests/test_buffer.cpp
tcw165: Right, that'd be simpler. I'll update the code.

src/tests/test_buffer.cpp
```diff
@@ -0,0 +1,156 @@
...
+            {
+                CHECK(mock.get_received_values().empty());
+                CHECK(mock.get_on_completed_count() == 0);
+                CHECK(mock.get_on_error_count() == 1);
+            }
+        }
+    }
+}
+
+SCENARIO("buffer bundles items", "[buffer][rvalue]")
```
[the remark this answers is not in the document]
tcw165: Oh it's just to make sure the rvalue overload works.

src/rpp/rpp/operators.hpp
```diff
@@ -35,6 +35,7 @@
  * \ingroup operators
  */
 
+#include <rpp/operators/buffer.hpp>
```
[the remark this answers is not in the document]
tcw165: You're right, and I'll move it.

src/rpp/rpp/operators/fwd/buffer.hpp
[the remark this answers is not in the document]
victimsnino: Example from RxPhp shows such an scenario: 

image

RxJava has next details: 

> the final emitted List may have fewer than count items

Anyway, i expect to see all emitted values somehow. For example, if buffer has size 100, but observable emitted 99 items, shouldn't i get it somehow?

src/rpp/rpp/operators/buffer.hpp
victimsnino: Your operator expects to obtain a subscriber of some specific type (let's say of `int`), but in case of obtaining subscriber of another type (let's say of `string`) you would accept some non-understandable issues related to inability to pass `string` where `int` expected. I think, it is better to have error like `expected subscriber of ints but obtained subscriber of strings`. Currently it is hard to obtain such an error, but anyway it improves overall readability of code

src/samples/doxygen/buffer.cpp
victimsnino: No, in such an case it would be also int =) because of `auto i = 0` will be deduced as int because of `0` is int. It would work in case of `auto i = size`, but not in this case

src/samples/doxygen/buffer.cpp
victimsnino: Oh, i see. But then it is better to place it inside tests, not doxygen sample. Because samples used as API reference and usage example for users. For me as for user it would be not clear what is difference in this case

src/tests/rxcpp_benchmark.cpp
victimsnino: as far as i remeber it should use `from`

src/tests/test_buffer.cpp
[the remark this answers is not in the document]
victimsnino: Actually, you have such an test at line 94 =) Observable returned from `flat_map` is rvalue

src/rpp/rpp/operators/buffer.hpp
tcw165: Do you have any recommended article giving more details?

src/rpp/rpp/operators/buffer.hpp
tcw165: Thanks for the explanation, and I'll update the code.

src/samples/doxygen/buffer.cpp
tcw165: You're right!

src/samples/doxygen/buffer.cpp
tcw165: Good point!

src/rpp/rpp/operators/buffer.hpp
tcw165: I found [this article](https://cplusplus.com/forum/beginner/261042/).

victimsnino: One more

https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/

src/rpp/rpp/operators/fwd/buffer.hpp
```diff
@@ -0,0 +1,80 @@
...
+private:
+    inline const SpecificObservable* CastThis() const
+    {
+        return static_cast<const SpecificObservable*>(this);
+    }
+
+    inline SpecificObservable&& MoveThis()
+    {
+        return std::move(*static_cast<SpecificObservable*>(this));
+    }
```
tcw165, line 78: For readability, I make these functions. But for performance, I inline them.

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,113 @@
...
+                auto copy = std::vector<DownstreamValueType>(shared_this->buckets_);
+                shared_this->buckets_.clear();
+                subscriber.on_next(std::move(copy));
+            }
+            subscriber.on_completed();
+        };
+    }
+
+private:
+    std::mutex mutex_;
```
victimsnino, line 87: You don't need mutex at all. Your operator doesn't provide any multi-threaded activity. But any previous operators MUST send you values in serialized way.

> Observables must issue notifications to observers serially (not in parallel). They may issue these notifications from different threads, but there must be a formal happens-before relationship between the notifications.
[source](https://reactivex.io/documentation/contract.html)

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,113 @@
...
+        {
+            std::lock_guard<std::mutex> guard(shared_this->mutex_);
+
+            shared_this->buckets_.push_back(value);
+            if (shared_this->buckets_.size() == shared_this->buckets_.capacity())
+            {
+                auto copy = std::vector<DownstreamValueType>(shared_this->buckets_);
+
+                shared_this->buckets_.clear();
+                subscriber.on_next(std::move(copy));
```
victimsnino, line 63: You can avoid extra copy
```suggestion
                subscriber.on_next(std::move(shared_this->buckets_));
                shared_this->buckets_.clear();
```

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,113 @@
...
+        return [shared_this = this->shared_from_this()](
+                const constraint::subscriber_of_type<DownstreamType> auto& subscriber)
+        {
+            std::lock_guard<std::mutex> guard(shared_this->mutex_);
+
+            if (!shared_this->buckets_.empty())
+            {
+                auto copy = std::vector<DownstreamValueType>(shared_this->buckets_);
+                shared_this->buckets_.clear();
+                subscriber.on_next(std::move(copy));
```
victimsnino, line 80: You can avoid extra copy
```suggestion
                subscriber.on_next(std::move(shared_this->buckets_));
                shared_this->buckets_.clear();
```

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+    template<constraint::subscriber_of_type<DownstreamType> TSub>
+    auto operator()(TSub&& subscriber) const
+    {
+        auto state = std::make_shared<buffer_on_next<UpstreamType, DownstreamType>>(count);
+        auto subscription = subscriber.get_subscription();
+
+        return create_subscriber_with_state<UpstreamType>(
+            std::move(subscription),
+            std::forward<decltype(subscriber)>(subscriber),
```
victimsnino, line 104:
```suggestion
            std::forward<TSub>(subscriber),
```

src/rpp/rpp/operators.hpp
```diff
@@ -56,6 +56,7 @@
...
```
victimsnino, line 59: Transforming operator, not combining =)

src/rpp/rpp/operators/fwd/buffer.hpp
victimsnino: I see, but `inline` is useless there. All functions placed inside `class/struct { ... }` by default are inline. So, you can remove it

victimsnino (changes requested):
Looks much better =)

src/rpp/rpp/operators/buffer.hpp
tcw165: Hmm, what if the `on_next()` in the middle of long chain in downstream raises an error. But then there's an `retry` operator appending after the downstream subscriber that raises the error. This observable gets resubscribed and the state here is not reset.

For the above reason, I thought it'd be better to reset the inner state and then propagate events to the downstream.

What do you think?

src/rpp/rpp/operators/buffer.hpp
tcw165: Let's follow the discussion on the other thread.

src/rpp/rpp/operators/buffer.hpp
victimsnino: Ok, let's check 2 situations:

1) (observable)-....-(buffer)-....-(retry)-....
in this case if any error between buffer and retry, than whole chain between observable and retry would be resubscribed and whole internal state recreated. For you it means that whole `buffer_on_next` will be recreated -> no any problems

2) (observable)-....-(retry)-.....-(buffer)-....
in this case if any error between observable and retry, than whole chain between observable and retry would be resubscribed, but buffer doesn't know about it  -> no any issues

src/rpp/rpp/operators/fwd/buffer.hpp
victimsnino, line 42:
```suggestion
            operator "buffer(2)" : +---{12}-{3}|
```

src/rpp/rpp/operators/buffer.hpp
tcw165: Thanks for clarification!

BTW, I think it's a typo in your code suggestion:

```
subscriber.on_next(std::move(shared_this->buckets_));
```

And this is what I think the right version:

```
subscriber.on_next(shared_this->buckets_);
```

We may still have a copy after `on_next()...` chain.

tcw165: Oh my bad, I just found that the `buckets_` will be cleared out after `std::move()`.

victimsnino: Ohhhh, it makes sense... Then yes, it is better to have some variable to store max size OR extract size before move and then make re-reserve

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+    // Copy & move assignment operators
+    buffer_on_next& operator=(const buffer_on_next&) = delete;
+    buffer_on_next& operator=(buffer_on_next&&) noexcept = default;
+
+    /// \return a on_next function for subscriber, that is copyable and all the
+    /// copies share the same state.
+    auto get_on_next() {
+        return [shared_this = this->shared_from_this()](
+                const UpstreamType& value,
```
victimsnino, line 51: Get value by `auto&&` and then forward it to vector to eliminate extra copy when move is possible

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,110 @@
...
+                const constraint::subscriber_of_type<DownstreamType> auto& subscriber)
+        {
+            shared_this->buckets_.push_back(value);
+            if (shared_this->max_ == 0 ||
```
victimsnino, line 55: max_ == 0 not possible due to your `max(1, count)` do only one thing

src/rpp/rpp/operators/buffer.hpp
```diff
@@ -0,0 +1,110 @@
...
+            if (!shared_this->buckets_.empty())
+            {
+                subscriber.on_next(std::move(shared_this->buckets_));
+            }
+            subscriber.on_completed();
+        };
+    }
+
+private:
+    inline void reserve_buckets() {
```
victimsnino, line 79: inline not needed

victimsnino (changes requested):
Last minor changes and can be merged =)

src/rpp/rpp/operators/buffer.hpp
tcw165: Let me try. I thought I had that `auto&&` iteration before but it didn't compile.

victimsnino (approved):

PR #190, opened 2022-07-27
SFML sample with simple snake game

victimsnino:


victimsnino:
@tcw165 , what do you think?

PR #199, opened 2022-08-01
Update test for "buffer" to use type not POD

tcw165:
In terms of "strong exception-safety guarantee", we'd like to keep track of the performance of "buffer" using non-POD types.

# Reference
- https://www.youtube.com/watch?v=3GwNjGMKBtI
- https://docs.microsoft.com/en-us/events/goingnative-2013/effective-cpp11-14-sampler

### DIFF
--- a/src/tests/benchmarks.cpp
+++ b/src/tests/benchmarks.cpp
@@ -547,13 +547,24 @@ TEST_CASE("concat")
 
 TEST_CASE("buffer")
 {
+    struct not_pod {
+        explicit not_pod(int i) : value(i) {};
+        not_pod(const not_pod&)                 = default;
+        not_pod(not_pod&&) noexcept             = default;
+
+        not_pod& operator=(const not_pod&)      = default;
+        not_pod& operator=(not_pod&&) noexcept  = default;
+
+        int value{0};
+    };
+
     BENCHMARK_ADVANCED("buffer")(Catch::Benchmark::Chronometer meter)
     {
-        auto sub = rpp::make_specific_subscriber<std::vector<int>>();
+        auto sub = rpp::make_specific_subscriber<std::vector<not_pod>>();
 
         meter.measure([&]
         {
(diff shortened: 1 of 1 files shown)


victimsnino:
Actually your `not_pod` is similar as pod. Due to `=default` for all constructors compiler would generate same efficient code as for POD. If you would like to test performance of buffer with class which doesn't follow "strong exception-safety guarantee", then you need to have move constructor which is not `noexcept`. Only in this case vector MAY produce less optimized code. Anyway, i'dont think we need it, due to we benchmark shows "maximum possible speed". Else we need to test, pods, pod-like, vector, string, big object, very big object, object without move constructor and etc

tcw165:
> Due to =default for all constructors compiler would generate same efficient code as for POD.

I can't delete those copy-* overload because somehow the compiler always omit the `on_next(T&&)` for `buffer()`. I'm quite confused by that and do you happen to know why?

tcw165:
> Anyway, i'dont think we need it, due to we benchmark shows "maximum possible speed"

Yep, I agree. Let's close this PR after you help me understand the issue above?

victimsnino:
> I can't delete those copy-* overload because somehow the compiler always omit the on_next(T&&) for buffer(). I'm quite confused by that and do you happen to know why?

Because `just` sends copy of values instead of moving original ones. So, add empty `.map([](auto v){return v;})` to convert copy to move

tcw165:
> Because just sends copy of values instead of moving original ones.

Can we refactor `just()` so that it sends moved values? I remember `std::move` for POD is very ok.

victimsnino:
No, it brokes it. Example

‘’’cpp
auto obs = rpp::source::just(std::vector{1});

mobs.subscribe(); // let’s imagine get by move


obs.subscribe(); // obtained empty vector????
‘’’


Only one ever way to do it: override const& and && for this source, but it increases complexity

PR #208, opened 2022-08-06
Add "trampoline" scheduler

tcw165:


src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include <chrono>
+#include <functional>
+#include <thread>
+
+namespace rpp::schedulers
+{
+
+/**
+ * \brief scheduler which schedules execution of schedulables via queueing tasks to the caller thread with priority to time_point and order
+ * \warning Creates new thread for each "create_worker" call. Any scheduled task will be queued to created thread for execution with respect to time_point and number of task
```
victimsnino, line 30: Incorrect description. Delete or change

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,102 @@
...
+                    }
+                }
+            } catch (const std::exception&) {
+                // no-op
+            }
+            queue_in_use = false;
+        }
+
+    private:
+        rpp::subscription_guard m_sub = rpp::subscription_base::empty();
```
victimsnino, line 76: what are the benefit of using `_guard` there? Do we really need to unsubscribe subscription when worker destructed?

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+        void defer_at(time_point time_point, std::invocable auto&& fn) const
+        {
+            m_state->defer_at(time_point, std::forward<decltype(fn)>(fn));
+        }
+
+        static time_point now() { return clock_type::now(); }
+
+    private:
+        std::shared_ptr<worker_strategy_state> m_state{};
```
victimsnino, line 94: Can't see if we really need shared_ptr here. composite_subscription is shared_ptr itself

src/tests/test_schedulers.cpp
```diff
@@ -357,6 +358,173 @@ SCENARIO("New thread scheduler depends on subscription")
...
+
+        worker.schedule([&]() -> rpp::schedulers::optional_duration
+                        {
+                            task_completed.set_value(std::this_thread::get_id());
+                            return std::nullopt;
+                        });
+
+        THEN("thread ID shall match with the current thread ID")
+        {
+            REQUIRE(task_future.wait_for(std::chrono::seconds{5}) == std::future_status::ready);
```
victimsnino, line 381: For this test we don't need to wait

src/tests/test_schedulers.cpp
```diff
@@ -357,6 +358,173 @@ SCENARIO("New thread scheduler depends on subscription")
...
+    auto scheduler = rpp::schedulers::trampoline{};
+    auto sub       = rpp::composite_subscription{};
+    auto worker    = scheduler.create_worker(sub);
+    rpp::subscription_guard guard{sub};
+
+    WHEN("supply a job that schedules inner job that schedules inner job")
+    {
+        std::vector<std::string> call_stack;
+
+        auto task_completed    = std::promise<void>{};
```
victimsnino, line 424: Vector of strings is enough. Don't overcomplicate test

src/tests/test_schedulers.cpp
```diff
@@ -357,6 +358,173 @@ SCENARIO("New thread scheduler depends on subscription")
...
+                    worker.schedule([&]() -> rpp::schedulers::optional_duration {
+                        call_stack.push_back(std::this_thread::get_id());
+                        call_stack_2.set_value(std::move(call_stack));
+                        return std::nullopt;
+                    });
+                    return std::nullopt;
+                });
+                return std::nullopt;
+            });
+        });
```
victimsnino, line 515: This two code blocks looks tooooooo clooooose. Extract some function to generate such an blocks and call twice (for thread 1 and for thread 2)

victimsnino (commented):
Mostly looks good

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,102 @@
...
+            queue.emplace(time_point, std::forward<decltype(fn)>(fn));
+            if (queue_in_use)
+                return;
+
+            try {
+                queue_in_use = true;
+                while (!queue.is_empty() && m_sub->is_subscribed())
+                {
+                    std::function<void()> schedulable{};
+                    if (queue.pop_if_ready(schedulable))
```
victimsnino, line 63: You can use pop_with_wait there due to anyway we MUST to wait for next schedulable

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
tcw165: You're right. There's no actual benefit since it's in sync with the current thread.

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
tcw165: I see. I'll reduce one layer of classes.

src/tests/test_schedulers.cpp
tcw165: In which file would you suggest for such utility function?

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,92 @@
...
+#include <rpp/subscriptions/composite_subscription.hpp> // lifetime
+#include <rpp/schedulers/details/queue_worker_state.hpp>// state
+
+#include <concepts>
+#include <chrono>
+#include <functional>
+#include <thread>
+
+thread_local bool queue_in_use = false;
+thread_local rpp::schedulers::details::queue_worker_state queue{};
```
tcw165, line 25: @victimsnino I found this is the easiest way to create thread-local queues but I'm not sure if it's the right way. Besides, I'm seeing compile error about duplicate symbols related to this.

src/samples/doxygen/trampoline.cpp
```diff
@@ -0,0 +1,34 @@
...
+//    rpp::source::just(rpp::schedulers::trampoline{}, 1, 2, 3)
+//        .combine_latest(rpp::source::just(rpp::schedulers::trampoline{}, 1, 2, 3))
+//        .observe_on(rpp::schedulers::trampoline{})
+//        .subscribe(
+//            [](const std::tuple<int, int> &v) { std::cout << v << "-"; },
+//            [](const std::exception_ptr &error) {},
+//            []() { std::cout << "|" << std::endl; });
+//    // Source: -1-2-3--|
+//    //         -1-2-3--|
+//    // Output: {1,1}-{2,2}-{3,3}-|
```
tcw165, line 31: @victimsnino I found the implementation of `just()` with the scheduler is probably unexpected:

```
        worker.schedule([iterable=std::forward<decltype(iterable)>(iterable), 
                         subscriber=std::forward<decltype(subscriber)>(subscriber), 
                         index = size_t{0}]() mutable-> schedulers::optional_duration
        {            
            try
            {
                const auto& extracted_iterable = extract_iterable_from_packed(iterable);
                const auto  end = std::cend(extracted_iterable);
                auto        itr = std::cbegin(extracted_iterable);

                std::ranges::advance(itr, static_cast<int64_t>(index), end);

                if (itr != end)
                {
                    subscriber.on_next(utils::as_const(*itr));
                    if (std::next(itr) != end) // it was not last
                    {
                        ++index;
                        return schedulers::duration{}; // re-schedule this
                    }
                }

                subscriber.on_completed();
            }
            catch(...)
            {
                subscriber.on_error(std::current_exception());
            }
            return std::nullopt;

        });
```

So there'd be only one schedulable that emits the elements one by one via `on_next()`. And that will lead to the result of `-{3,1}-{3,2}-{3,3}-|`.

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
victimsnino: > I'm seeing compile error about duplicate symbols related to this

Yes, because you've created global symbols and each compilation unit which includes this would try to create this one. Place them inside scheduler body

src/tests/test_schedulers.cpp
victimsnino: You can extract it into same file. Or even create lambda inside a test, it doesn't matter

src/samples/doxygen/trampoline.cpp
victimsnino: Probably i didn't catch your point. Can't see if anything is wrong? We need to schedule each item emission and send only one value per scheduling. Expected result would be something like this
```
> inner observable send 1
> outer observable send 1
{1,1}
> inner observable send 2
{1,2}
>outer observable send 2
{2,2}
> inner observable send 3
{2, 3}
> outer observable send 3
{3,3}
```

src/rpp/rpp/schedulers/details/queue_worker_state.hpp
```diff
@@ -74,7 +74,7 @@ class queue_worker_state
         return false;
     }
 
-    void reset()
+    void reset() noexcept
```
victimsnino, line 77: Don't overcomplicate and avoid trying to do anything as noexcept =) This function is pretty simple and can be inlined by compiler without any tips

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+namespace rpp::schedulers
+{
+
+template <typename F>
+struct scope_guard {
+    explicit scope_guard(F&& f) : m_f(std::forward<F>(f)) {}
```
victimsnino, line 32: You need move instead of forward there. `&&` is forwarding **ONLY** when you have template argument for this one FOR function.
```cpp
template <typename F>
struct scope_guard {
    template<typename U>
    explicit scope_guard(U&& f) : m_f(std::forward<U>(f)) {}
```
^^^^^ is forwarding reference due to it is deduced FROM argument of function (constrctor). But in your case you class specialized with template and then you accept it by **rvalue-reference**

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,92 @@
...
+            queue_in_use = true;
+            scope_guard cleanup([]() noexcept
+            {
+                // If error occurs, we still want the thread-local state to be reset.
+                queue_in_use = false;
+            });
+            while (!queue.is_empty() && m_sub.is_subscribed())
+            {
+                std::function<void()> schedulable{};
+                std::stop_token stop;
```
victimsnino, line 72: stop token used to unblock wait in case of someone requested stop. Add it to `m_sub` to request stop in case of unsubscribe

src/samples/doxygen/trampoline.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * \example trampoline.cpp
+ **/
+int main()
+{
+    //! [trampoline]
+    rpp::source::just(rpp::schedulers::trampoline{}, 1, 2, 3)
+        .merge_with(rpp::source::just(rpp::schedulers::trampoline{}, 4, 5, 6))
+        .observe_on(rpp::schedulers::trampoline{})
```
victimsnino, line 12: I think we don't need it because it is overcomplication. Change to subscribe_on

src/tests/test_schedulers.cpp
```diff
@@ -357,6 +358,187 @@ SCENARIO("New thread scheduler depends on subscription")
     }
 }
 
+SCENARIO("trampoline scheduler dispatches task in the same thread")
+{
+    auto scheduler = rpp::schedulers::trampoline{};
+    auto sub       = rpp::composite_subscription{};
+    auto worker    = scheduler.create_worker(sub);
+    rpp::subscription_guard guard{sub};
```
victimsnino, line 366: ???? Can't see why do you need it

src/tests/test_schedulers.cpp
```diff
@@ -357,6 +358,187 @@ SCENARIO("New thread scheduler depends on subscription")
...
+
+    WHEN("supply a simple job")
+    {
+        auto task_completed    = std::promise<std::thread::id>{};
+        auto task_future       = task_completed.get_future();
```
victimsnino, line 371: std::optional would be enough =)

src/samples/doxygen/trampoline.cpp
tcw165: Oh I meant that I have this code and this output:

Code:
```
    rpp::source::just(rpp::schedulers::trampoline{}, 1, 2, 3)
        .with_latest_from(rpp::source::just(rpp::schedulers::trampoline{}, 4, 5, 6))
        .subscribe(
            [](const auto &v) { std::cout << v << "-"; },
            [](const std::exception_ptr &error) {},
            []() { std::cout << "|" << std::endl; });```

Output

```
-{1,6}-{2,6}-{3,6}-|
```

And the reason for that is because we only have one master schedulable that handles all the `on_next()`. Alternatively, we may consider creating new schedulable for each `on_next()`?

File: src/rpp/rpp/schedulers/trampoline_scheduler.hpp (hunk 11 above)
Comment (tcw165) [reply src/rpp/rpp/schedulers/trampoline_scheduler.hpp:32]: You're right. Thanks for catching this :)

File: src/rpp/rpp/schedulers/trampoline_scheduler.hpp (hunk 12 above)
Comment (tcw165) [reply src/rpp/rpp/schedulers/trampoline_scheduler.hpp:72]: Yup, I have the same question too. I'll add that later.

File: src/samples/doxygen/trampoline.cpp (hunk 9 above)
Comment (victimsnino) [reply src/samples/doxygen/trampoline.cpp]: In this case you need subscribe_on. I will fix after merging your PR due to it is required extra code in subscribe method

Comment (tcw165) [reply src/samples/doxygen/trampoline.cpp]: Please let me know about the PR after this is merged.

File: src/rpp/rpp/schedulers/trampoline_scheduler.hpp [hunk 16]
```diff
@@ -0,0 +1,101 @@
...
+                    schedulable = {};
+                }
+            }
+        }
+
+        static time_point now() { return clock_type::now(); }
+
+    private:
+        rpp::composite_subscription m_sub;
+        std::shared_ptr<std::stop_source> m_stop;
```
victimsnino, line 93: you don't need it as shared_ptr. You can create it during `defer_at` function. Add callback to subscription and remove/unsubscribe it after while loop to remove local variable from m_sub state

src/tests/test_schedulers.cpp
```diff
@@ -357,6 +384,176 @@ SCENARIO("New thread scheduler depends on subscription")
...
+        THEN("thread ID shall match with the current thread ID")
+        {
+            REQUIRE(observed_thread_id.value() == std::this_thread::get_id());
+        }
+    }
+
+    WHEN("supply a job that schedules inner job")
+    {
+        auto task_completed    = std::promise<std::thread::id>{};
+        auto task_future       = task_completed.get_future();
```
victimsnino, line 412: Same as above -> use optional

src/tests/test_schedulers.cpp
```diff
@@ -357,6 +384,176 @@ SCENARIO("New thread scheduler depends on subscription")
...
+        auto scheduler = rpp::schedulers::trampoline{};
+        auto sub       = rpp::composite_subscription{};
+        auto worker    = scheduler.create_worker(sub);
+
+        THEN("shall not see execution of inner schedulable")
+        {
+            std::jthread async_stopper([=] {
+                std::this_thread::sleep_for(std::chrono::seconds{1});
+                sub.unsubscribe();
+            });
```
victimsnino, line 540: Can't see any benefit of using thread there comparing to just calling unsubscribe in place of 
```cpp
// Block return so that async stopper could prevent the inner schedulable from running.
std::this_thread::sleep_for(std::chrono::seconds{3});
```

victimsnino: Threads and sleeps add complexity to test, but doesn't provide any benefit and just adds race-condition.

tcw165: You're right! I've been having this concern for long time.

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
tcw165: I used a shared pointer for shared ownerships. Do you mean something like below in `defer_at()`?

```
std::stop_source stop;
auto stop_token = stop.get_token();

m_sub.add([stop = std::move(stop)] { // move stop since subscription callback is the sole owner.
  stop.request_stop();
});

while (...) {
  if (queue.pop_wait(..., stop_token) ...
}
```

tcw165: I feel the `stop_source` has the same lifetime to `m_sub` and thus I lift it as a class member.

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
tcw165: Discussed offline: we use a static class function to house static `thread_local` variables.

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,117 @@
...
+#include <functional>
+#include <thread>
+
+
+namespace rpp::schedulers
+{
+
+template <typename F>
+struct scope_guard {
+    explicit scope_guard(F&& f) : m_f(std::move<F>(std::forward<F>(f))) {}
```
victimsnino, line 30: Use forward ONLY when template parameter is forwarding reference. In your case it is  rvalue
```suggestion
    explicit scope_guard(F&& f) : m_f(std::move<F>(f)) {}
```

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,117 @@
...
+    public:
+        explicit worker_strategy(const rpp::composite_subscription& subscription)
+            : m_sub{subscription} {};
+
+        void defer_at(time_point time_point, std::invocable auto&& fn) const
+        {
+            if (!m_sub.is_subscribed())
+                return;
+
+            with_thread_local_schedulable_queue([&](auto& queue, auto& queue_in_use)
```
victimsnino, line 57: It would be more understandable =)
```suggestion
            with_thread_local_schedulable_queue([&](auto& queue, bool& queue_in_use)
```

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,117 @@
...
+                {
+                    // If error occurs, we still want the thread-local state to be reset.
+                    queue_in_use = false;
+                    if (!m_sub.is_subscribed())
+                        queue.reset();
+                });
+
+                std::stop_source stop;
+                auto stop_token = stop.get_token();
+                m_sub.add([stop = std::move(stop)]()
```
victimsnino, line 75: looks like you need to mark it as mutable to be compiled

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -0,0 +1,117 @@
...
+        return worker<worker_strategy>{sub};
+    }
+
+private:
+    /**
+     * \brief evaluates given lambda with the thread local queue and ownership-flag.
+     *
+     * \param action a lambda that is given with the thread local queue and ownership-flag.
+     */
+    static void with_thread_local_schedulable_queue(std::function<void(rpp::schedulers::details::queue_worker_state&, bool&)>&& action)
```
victimsnino, line 110: std::function is complex and heavy type. You can use auto there or something like `const std::invocable<rpp::schedulers::details::queue_worker_state&, bool&> auto&`

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
victimsnino:
```suggestion
                m_sub.add([stop = std::move(stop)]() mutable
```

victimsnino (approved):

PR #216, opened 2022-08-11
[WIP] Update trampoline scheduler

victimsnino:
## Changes:
- refactor code to make it faster and more structural
- speedup trampoline scheduler
- use it inside specific_observable to take ownership over queue
- get rid of recursion in immediate scheduler

[WIP]:
- need to improve performance via elimination of copy of std::function
- need to improve performance via using "immediate-like" scheduling till it is only one schedulable
- make wait stoppable

### DIFF
--- a/src/rpp/rpp/observables/specific_observable.hpp
+++ b/src/rpp/rpp/observables/specific_observable.hpp
@@ -10,13 +10,13 @@
 
 #pragma once
 
+#include <rpp/defs.hpp>
 #include <rpp/observables/fwd.hpp>
 #include <rpp/observables/interface_observable.hpp>            // base_class
-#include <rpp/utils/operator_declaration.hpp>                  // for header include
+#include <rpp/schedulers/trampoline_scheduler.hpp>
 #include <rpp/subscribers/dynamic_subscriber.hpp>
+#include <rpp/utils/operator_declaration.hpp>                  // for header include
 #include <rpp/utils/utilities.hpp>                            // copy_assignable_callable
-
-#include <rpp/defs.hpp>
 
 #include <utility>
 
@@ -71,6 +71,8 @@ private:
     {
         try
         {
+            // take ownership over current thread as early as possible to delay all next "current_thread" schedulings. For  example, scheduling of emissions from "just" to delay it till whole chain is subscribed and ready to listened emissions
+            const auto drain_handle = rpp::schedulers::current_thread::ensure_queue_if_no_any_owner();
             m_state(subscriber);
         }
         catch (...)
--- a/src/rpp/rpp/observables/specific_observable.hpp
+++ b/src/rpp/rpp/observables/specific_observable.hpp
@@ -72,8 +72,18 @@ private:
         try
         {
             // take ownership over current thread as early as possible to delay all next "current_thread" schedulings. For  example, scheduling of emissions from "just" to delay it till whole chain is subscribed and ready to listened emissions
-            const auto drain_handle = rpp::schedulers::current_thread::ensure_queue_if_no_any_owner();
-            m_state(subscriber);
+            if (rpp::schedulers::current_thread::is_queue_owned())
+            {
+                m_state(subscriber);
+            }
+            else
+            {
+                // will be scheduled immediately -> reference can be passed
+                rpp::schedulers::current_thread::create_worker(subscriber.get_subscribtion()).schedule([&]
+                {
+                    m_state(subscriber);
+                });
+            }
         }
         catch (...)
         {
--- a/src/rpp/rpp/observables/specific_observable.hpp
+++ b/src/rpp/rpp/observables/specific_observable.hpp
@@ -72,6 +72,10 @@ private:
         try
         {
             // take ownership over current thread as early as possible to delay all next "current_thread" schedulings. For  example, scheduling of emissions from "just" to delay it till whole chain is subscribed and ready to listened emissions
+            // For example, if we have
+            // rpp::source::just(rpp::schedulers::current_thread{}, 1,2).combine_latest(rpp::source::just(rpp::schedulers::current_thread{}, 1,2))
+            //
+            // then we expect to see emissions like (1,1) (2,1) (2,2) instead of (2,1) (2,2). TO do it we need to "take ownership" over queue to prevent ANY immediate schedulings from ANY next subscriptions
             if (rpp::schedulers::current_thread::is_queue_owned())
             {
                 m_state(subscriber);
--- a/src/rpp/rpp/observables/specific_observable.hpp
+++ b/src/rpp/rpp/observables/specific_observable.hpp
@@ -79,9 +79,10 @@ private:
             else
             {
                 // will be scheduled immediately -> reference can be passed
-                rpp::schedulers::current_thread::create_worker(subscriber.get_subscribtion()).schedule([&]
+                rpp::schedulers::current_thread::create_worker(subscriber.get_subscription()).schedule([&]
                 {
                     m_state(subscriber);
+                    return rpp::schedulers::optional_duration{};
                 });
             }
         }
--- a/src/rpp/rpp/schedulers/details/queue_worker_state.hpp
+++ b/src/rpp/rpp/schedulers/details/queue_worker_state.hpp
@@ -19,6 +19,7 @@
 
 namespace rpp::schedulers::details
 {
+template<typename SchedulableFn>
 class schedulable
 {
 public:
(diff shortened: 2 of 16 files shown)


victimsnino:
@tcw165, you wanted to check it

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -96,22 +121,22 @@ class trampoline final : public details::scheduler_tag
...
+    [[nodiscard]] static utils::finally_action<void(*)()> ensure_queue_if_no_any_owner()
     {
-        return worker<worker_strategy>{sub};
+        auto& queue = get_schedulable_queue();
+        if (queue.has_value())
+            return utils::finally_action{+[] {}};
+
+        queue = std::priority_queue<current_thread_schedulable>{};
+        return utils::finally_action{&drain_queue};
     }
```
tcw165, line 135: nit: For better readability, would you mind to move this function to closer to `wait_and_extract_executable_if_subscribed` at top?

src/rpp/rpp/utils/utilities.hpp
```diff
@@ -85,4 +85,25 @@ class copy_assignable_callable
...
+
+/**
+ * \brief Calls passed function during destruction
+ */
+template<std::invocable Fn>
+class finally_action
+{
+public:
+    finally_action(Fn&& fn)
+        : m_fn{std::move(fn)} {}
```
tcw165, line 97: For my education, why don't you need a `std::forward()` inside the `std::move()`? Is it bc move just cast whatever given value to rvalue reference?

src/rpp/rpp/schedulers/details/queue_worker_state.hpp
```diff
@@ -18,6 +18,33 @@
...
+    schedulable& operator=(const schedulable& other)     = default;
+    schedulable& operator=(schedulable&& other) noexcept = default;
+
+    bool operator<(const schedulable& other) const
+    {
+        return std::tie(m_time_point, m_id) >= std::tie(other.m_time_point, other.m_id);
+    }
+
+    time_point              GetTimePoint() const { return m_time_point; }
+    std::function<void()>&& ExtractFunction() const { return std::move(m_function); }
```
tcw165, line 40: nit: Why are these two functions in Camel Case (`SamIsAwesome`) not Snake Case (`sam_is_awesome`)?

src/rpp/rpp/schedulers/details/queue_worker_state.hpp
```diff
@@ -18,6 +18,33 @@
...
+
+private:
+    time_point                    m_time_point;
+    size_t                        m_id;
+    mutable std::function<void()> m_function;
```
tcw165, line 45: > To eliminate `std::function`

This `std::function` is needed and inevitable right?

src/rpp/rpp/schedulers/immediate_scheduler.hpp
```diff
@@ -27,32 +27,48 @@ namespace rpp::schedulers
...
+            schedule(now() + delay, std::forward<decltype(fn)>(fn));
+        }
 
-            if (m_sub.is_subscribed())
-                fn();
+        void schedule(time_point time_point, constraint::schedulable_fn auto&& fn) const
+        {
+            while (m_sub.is_subscribed())
+            {
+                std::this_thread::sleep_until(time_point);
```
tcw165, line 50: nit: Not sure if we want stoppable sleep?

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -42,52 +36,83 @@ struct scope_guard {
...
+                function();
+        }
+    }
+
+    [[nodiscard]] static std::function<void()> wait_and_extract_executable_if_subscribed(const  current_thread_schedulable& schedulable)
+    {
+        if (!schedulable.is_subscribed())
+            return {};
+
+        std::this_thread::sleep_until(schedulable.GetTimePoint());
```
tcw165, line 94: The `std::this_thread::sleep_until(schedulable.GetTimePoint());` is un-stoppable vs the stoppable `pop_with_wait(..., token)` in the previous implementation.
There could be this case the sharable subscription is held in other thread and somehow the subscription is unsubscribed in other thread. With the `::sleep_until()`, we can’t stop the thread responsively but have to wait for the thread to wake up and quit.

Can we use stoppable sleeping mechanism here?

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
tcw165: nvm, it's a **public** static method.

src/rpp/rpp/observables/specific_observable.hpp
```diff
@@ -71,6 +71,8 @@ class specific_observable : public interface_observable<Type, specific_observabl
     {
         try
         {
+            // take ownership over current thread as early as possible to delay all next "current_thread" schedulings. For  example, scheduling of emissions from "just" to delay it till whole chain is subscribed and ready to listened emissions
+            const auto drain_handle = rpp::schedulers::current_thread::ensure_queue_if_no_any_owner();
```
tcw165, line 75: For everyone's education, can you put a comment with the stack iteration for the very reason of the drain here?

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
```diff
@@ -42,52 +36,83 @@ struct scope_guard {
...
-                    std::function<void()> schedulable{};
-                    if (queue.pop_with_wait(schedulable, stop_token))
-                    {
-                        schedulable();
-                        schedulable = {};
-                    }
-                }
-            });
+            auto drain_handle = ensure_queue_if_no_any_owner();
+            get_schedulable_queue()->emplace(time_point, std::forward<decltype(fn)>(fn), m_sub);
```
tcw165, line 115: I understand the needs for reuse of `ensure_queue_if_no_any_owner()`. But here is a concern in terms of the readability.

The code was flat and the control flow is explained in one place. But with calling nested function here, I'm worried the readers will need stronger background knowledge for behavior of **trampoline** to read the code. For example:

1. Readers need to dive into the `ensure_queue_if_no_any_owner()` to know it's using RAII to dequeuing the message queue in the current thread.
2. Readers need to realize that we enqueue the task/schedulable before #1 really happens.

tcw165: > Readers need to dive into the `ensure_queue_if_no_any_owner()` to know it's using RAII to dequeuing the message queue in the current thread.

You may want to add extra comment in `ensure_queue_if_no_any_owner()` for the reason why executing on destructing is important.

tcw165: Another thought:

To eliminate `std::function` from my initial implementation, `with_thread_local_schedulable_queue()`, we could use `invocable` concept and feed a member function to it.

src/rpp/rpp/utils/utilities.hpp
victimsnino: Okay, let's start again =)))
```cpp
struct A
{
// it just function that accepts just int. It would be copied or moved when function called
void func(int){}
};

struct A
{
// it is rvalue-reverence. int MUST be moved when function called
void func(int&&){}
};

// It is template class. So, when you construct A you MUST to specify it with some SPECIFIC type. For example, if A<int>, then func accepts int via copy/move. If A<int&&>, then rvalue  reference-> must be moved in.
template<typename T>
struct A
{
   void func(T) {}
}

// same as before, but func REQUIRES to accept T by rvalue reference in any case. So, A<int> and A<int&&> are same for us
template<typename T>
struct A
{
   void func(T&&) {}
}

struct A
{
// it just function that accepts just any type. It would be copied or moved when function called. Except of specializing type manually
template<typename T>
void func(T){}
}

struct A
{
// it just function that accepts just any type by reference
template<typename T>
void func(const T&){}
}

struct A
{
// it just function that accepts just any type by forwarding/universal/perfect reference. So, final type of T would be deduced based on passed object. if you pass func(int{}), then T is just int. if you pass func(my_variable), then T is just int& and references collapsing rule collapse it to just reference.
template<typename T>
void func(T&&){}
}

template<typename T>
struct A
{
// same as before
template<typename U>
void func(U&&){}
}
```

victimsnino: As a result, in this case it just accepts RVALUE to Fn, not forwarding reference

src/rpp/rpp/schedulers/details/queue_worker_state.hpp
victimsnino: This one - yes

src/rpp/rpp/schedulers/trampoline_scheduler.hpp
tcw165: Thanks a lot ❤️

victimsnino: Can you check if it come more understandable?

PR #218, opened 2022-08-14
Add "combine_latest" operator

tcw165:


src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,169 @@
...
+                coordinator->offer_completed(subscriber);
+            }));
+    }
+
+public:
+    template<constraint::subscriber_of_type<DownstreamType> TSub>
+    auto operator()(TSub&& subscriber) const
+    {
+        // TODO: Avoid copying the combiner?
+        auto coordinator = std::make_shared<combine_latest_coordinator_t<Type, utils::extract_observable_type_t<TObservable>...>>(m_combiner);
```
tcw165, line 147: @victimsnino I make `combine_latest_coordinator_t` private class to this class so that it's hidden from the outer world.

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,173 @@
...
+        auto on_next = [coordinator](auto&& value, const auto& subscriber)
+        {
+            coordinator->template offer_next_value<INDEX_OF_SOURCE_TYPE>(std::forward<decltype(value)>(value), subscriber);
+        };
+        auto on_completed = [coordinator](const auto& subscriber)
+        {
+            coordinator->offer_completed(subscriber);
+        };
+
+        return create_subscriber_with_state<Type>(std::forward<TSub>(subscriber),
```
victimsnino, line 166:
```suggestion
        auto sub = subscriber.get_subscription().make_child(),
        return create_subscriber_with_state<Type>(std::move(sub),
                                                  std::forward<TSub>(subscriber),
```

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,173 @@
...
+            subscriber,
+            [coordinator](auto&& value, const auto& subscriber)
+            {
+                coordinator->template offer_next_value<I>(std::forward<decltype(value)>(value), subscriber);
+            },
+            utils::forwarding_on_error{},
+            [coordinator](const auto& subscriber)
+            {
+                coordinator->offer_completed(subscriber);
+            }));
```
victimsnino, line 143: Subscriber there and inside `operator()` are same. Extract function for creation of subscriber and call it there and inside `operator()`

victimsnino (changes requested):
Some changes from brief overview

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,173 @@
...
+     * \param subscriber is the downstream subscriber.
+     * \param coordinator manages the cache of emission from the observables and coordinate dispatching.
+     */
+    template<size_t...I>
+    static void subscribe_other_observables(std::index_sequence<I...>, // Used in compile time for variadic expansion
+                                            const auto& observables,
+                                            const auto& subscriber,
+                                            const auto& coordinator)
+    {
+        // +1 because the first element in tuple is the current observable, and you want to subscribe to the 'other' observables.
```
victimsnino, line 123: It is not true, because tuple contains only "other" observables, while subscribe on "current" happens inside `operator()` via returning subscriber for "current"

tcw165: I don't quite understand. I design the tuple to include this observable and other observable.

src/rpp/rpp/operators/combine_latest.hpp
tcw165: In what cases we need to create a children subscription?

src/rpp/rpp/operators/combine_latest.hpp
tcw165: Oh can you provide me more snippet? I needed this function for the variadic expansion for `template<size_t...I>`.

src/rpp/rpp/operators/combine_latest.hpp
victimsnino: Discussed on gitter

src/rpp/rpp/operators/combine_latest.hpp
victimsnino:
```cpp
template<typename ValueType, size_t I>
auto create_inner_subscriber(auto&& subscriber, const auto& coordinator)
{
auto sub = subscriber.get_subscription().make_child();
return create_subscriber_with_state<ValueType>(
            std::move(sub),
            std::forward<decltype(subscriber)>(subscriber),
            [coordinator](auto&& value, const auto& subscriber)
            {
                coordinator->template offer_next_value<I>(std::forward<decltype(value)>(value), subscriber);
            },
            utils::forwarding_on_error{},
            [coordinator](const auto& subscriber)
            {
                coordinator->offer_completed(subscriber);
            })
}


 static void subscribe_observable(const TObs& observable, const auto& subscriber, const auto& coordinator)
    {
        using ValueType = utils::extract_observable_type_t<TObs>;
        observable.subscribe(create_inner_subscriber<ValueType, I>(subscriber, coordinator));
    }


public:

template<constraint::subscriber_of_type<DownstreamType> TSub>
    auto operator()(TSub&& subscriber) const
    {
        // TODO: Avoid copying the combiner?
        auto coordinator = std::make_shared<combine_latest_coordinator_t<Type, utils::extract_observable_type_t<TObservable>...>>(m_combiner);

        // Subscribe to other observables and redirect on_next event to coordinator
        subscribe_other_observables(std::index_sequence_for<TObservable...>{}, m_other_observables, subscriber, coordinator);

       return create_inner_subscriber<ValueType, INDEX_OF_SOURCE_TYPE>(std::forward<TSub>(subscriber), coordinator);
    }
```

src/rpp/rpp/operators/combine_latest.hpp
victimsnino: Oh, I see, my miss-understanding... But it looks really hard to understand =)

tcw165: Haha, my bad. I know it's hard especially the behavior of this is quite similar to `with_latest_from()`. But it does make sense to have a tuple for this observable + other observables.

src/rpp/rpp/operators/combine_latest.hpp
tcw165: Hmm, I don't think the code above will work. Can we keep the current code?

tcw165: Oh can I keep this pattern for better readability? Because not all the IDE(s) tell you the argument's name, it's kind of readable by having a variable name. Also, I believe all the compilers will optimize it by lifting the variable as unnamed variable?

```
auto on_next = [] {};
auto on_completed = [] {};

return create_subscriber_with_state<ValueType>(
    ...,
    std::move(on_next),
    ...,
    std::move(on_completed));
```

victimsnino: But why? your code contains long duplicate and it is better to extract this into separate function to show that ALL inner subscribers are same

victimsnino: > Hmm, how does the variadic expansion work for the variadic observables (this one + others)?

Not sure if i get your question, there is no any variadic...

victimsnino: Whole this comment about extracting `create_inner_subscriber` as separate function

victimsnino: > Oh can I keep this pattern for better readability? Because not all the IDE(s) tell you the argument's name, it's kind of readable by having a variable name. Also, I believe all the compilers will optimize it by lifting the variable as unnamed variable?

Not sure if is really needed. Pattern of creation of all operators are the same. Same about creation of inner subscribers - subscribtion/state/on_next/on_error/on_completed. Moreover, it is fully internal thing and some "non-experienced" user not need to read this code, while some "experienced" can understand it without named variables

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,166 @@
...
+ * \brief "combine_latest" operator (an OperatorFn used by "lift").
+ */
+template<constraint::decayed_type Type, typename TCombiner, constraint::observable ...TOtherObservable>
+struct combine_latest_impl
+{
+    TCombiner m_combiner;
+    std::tuple<TOtherObservable...> m_other_observables;
+
+private:
+    static constexpr size_t INDEX_OF_SOURCE_TYPE = 0;
```
victimsnino, line 39:
```suggestion
    static constexpr size_t s_index_of_source_type = 0;
```

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+    using DownstreamType = utils::decayed_invoke_result_t<TCombiner, Type, utils::extract_observable_type_t<TOtherObservable>...>;
+
+    /**
+     * \brief The coordinator that caches the values from all the observables and dispatches latest caches to the observer. Note the emission is only sent to the observer when all the observables at least emits once.
+     */
+    template<constraint::decayed_type SourceValueType, constraint::decayed_type ...OtherValueTypes>
```
victimsnino, line 46: we don't need these templates due to all of them can be deduced from templates of `combine_latest_impl`. But extra templates increases compilation time/binary size

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,166 @@
...
+    struct combine_latest_coordinator_t
+    {
+        explicit combine_latest_coordinator_t(TCombiner combiner) : m_combiner(combiner) {}
```
victimsnino, line 49:
```suggestion
        explicit combine_latest_coordinator_t(const TCombiner& combiner) : m_combiner(combiner) {}
```

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+        // Not copyable nor movable.
+        combine_latest_coordinator_t(const combine_latest_coordinator_t&) = delete;
+        combine_latest_coordinator_t(combine_latest_coordinator_t&&) = delete;
+
+        template<size_t I>
+        void offer_next_value(auto&& value, const auto& subscriber)
+        {
+            if (!subscriber.is_subscribed()) return;
```
victimsnino, line 58: don't needed

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,166 @@
...
+            // If all the optional(s) have value, dispatch to downstream
+            if (has_values_for_observer())
+            {
+                subscriber.on_next(combine_cached_values());
+            }
+        }
+
+        void offer_completed(const auto& subscriber)
+        {
+            if (!subscriber.is_subscribed()) return;
```
victimsnino, line 72: Don't need

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+            std::scoped_lock<std::mutex> lock{m_mutex};
+
+            if (++m_completed_count == m_total_completed_count)
```
victimsnino, line 76: we don't need mutex there. Can be changed to atomic_size_t

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,166 @@
...
+        std::tuple<std::optional<SourceValueType>, std::optional<OtherValueTypes>...> m_values{};
+        size_t m_completed_count{0};
+
+        const std::size_t m_total_completed_count = 1 + sizeof...(OtherValueTypes);
+
+        bool has_values_for_observer()
+        {
+            return std::apply([](const auto&... cached_values)
+            {
+                // Variadic expansion used
```
victimsnino, line 95: not sure if this comment needed =)

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,166 @@
...
+                return (cached_values.has_value() && ...);
+            }, m_values);
+        }
+
+        auto combine_cached_values()
+        {
+            return std::apply([&](const auto&... cached_values)
+            {
+                // Variadic expansion used
```
victimsnino, line 104: not sure if this comment needed =)

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * \brief Templated helper function for subscribing to variadic 'other' observables.
+     *
+     * \param observables is a variadic tuple. The first observable is the current observable and rest are the other observables.
+     * \param subscriber is the downstream subscriber.
+     * \param coordinator manages the cache of emission from the observables and coordinate dispatching.
+     */
+    template<size_t...I>
+    static void subscribe_other_observables(std::index_sequence<I...>, // Used in compile time for variadic expansion
```
victimsnino, line 118: Why static? we pass m_other_observables as observables, but it can be just "member non-static function"

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -0,0 +1,166 @@
...
+        using ValueType = utils::extract_observable_type_t<TObservable>;
+        observable.subscribe(create_inner_subscriber<ValueType, I>(subscriber, coordinator));
+    }
+
+    template<typename ValueType, size_t I>
+    static auto create_inner_subscriber(auto&& subscriber, const auto& coordinator)
+    {
+        return create_subscriber_with_state<ValueType>(
+                    subscriber.get_subscription().make_child(),
+                    std::forward<decltype(subscriber)>(subscriber),
```
victimsnino, line 140: Error-prone line! extract getting subscription as separate line and then move it there. What if subscriber was moved via `forward` before getting subscription???

src/tests/benchmarks.cpp
```diff
@@ -391,6 +391,37 @@ TEST_CASE("with_latest_from")
...
+{
+    BENCHMARK_ADVANCED("combine_latest construction from observable via dot + subscribe")(Catch::Benchmark::Chronometer meter)
+    {
+        const auto obs_1 = rpp::source::just(1, 2, 3);
+        const auto obs_2 = rpp::source::just(4, 5, 6);
+        auto sub = rpp::specific_subscriber{[](const std::tuple<int, int>) {}};
+
+        meter.measure([&]
+        {
+            return obs_1.combine_latest(obs_2).subscribe(sub);
```
victimsnino, line 404: You need vector of subs there. Because `measure` can invoke lambda multple times. In this case first launch "completes" sub and then you measure nothing (due to sub unsubscribed). Re-use code with creation of vector of subs from above

src/tests/benchmarks.cpp
```diff
@@ -391,6 +391,37 @@ TEST_CASE("with_latest_from")
...
+        });
+    };
+
+    BENCHMARK_ADVANCED("sending of values from observable via combine_latest to subscriber")(Catch::Benchmark::Chronometer meter)
+    {
+        auto inner_source = rpp::source::just(1);
+        using inner_source_type = decltype(inner_source);
```
victimsnino, line 411: You can send just int?

victimsnino: Not sure if you need inner source there

victimsnino (changes requested):

tcw165:
@victimsnino Can you revisit the PR?

src/rpp/rpp/operators/combine_latest.hpp
tcw165: Oh just a friendly reminder for someone who doesn't know about this and will become "experienced" over time.

src/rpp/rpp/operators/combine_latest.hpp
tcw165: Oh you're right. I was thinking if this will work last night too.

src/rpp/rpp/operators/combine_latest.hpp
tcw165: Oh I see.

src/rpp/rpp/operators/combine_latest.hpp
tcw165: Oh this is ok bc the `combine_latest_impl` lives longer than `combine_latest_coordinator_t`?

src/rpp/rpp/operators/combine_latest.hpp
tcw165: I understand you want lock-free monitoring on completed event. But using `atomic_size_t` will introduce a second mutex, which I think it's more expensive than having other `on_next()` wait on the same mutex.

What do you think?

src/tests/benchmarks.cpp
tcw165: Oh I see. This explains why you have `create()` upstream that never ends.

src/tests/benchmarks.cpp
tcw165: I see your point and I'll update them.

src/rpp/rpp/operators/combine_latest.hpp
tcw165: > But using atomic_size_t will introduce a second mutex

It may not be true for all the platform. You'll have better experience to guide me.

victimsnino: Atomic size_t should be lock_free as far as i know. Anyway, it would be not slower than mutex, but in most cases faster

src/rpp/rpp/operators/combine_latest.hpp
victimsnino: No, the order of evaluation of function arguments is unspecified. As a result, std::forward can happens BEFORE getting of the subscription

victimsnino: Simple example on [godbolt](https://godbolt.org/z/Mbv9WEodK)

```cpp
#include <iostream>

struct A
{
    A(int i) { std::cout << "Constructed A " << i << std::endl; }
};

void f(A a1, A a2, A a3){}

int main()
{
    f(A{1}, A{2}, A{3});
    return 0;
}
```
MSVC/GCC
```
Constructed A 3
Constructed A 2
Constructed A 1
```
Clang
```
Constructed A 1
Constructed A 2
Constructed A 3
```

tcw165: Wow, this is really interesting!

src/rpp/rpp/operators/combine_latest.hpp
tcw165: Agree, I just patched the code.

src/rpp/rpp/operators/fwd/combine_latest.hpp
```diff
@@ -0,0 +1,125 @@
...
+template<constraint::decayed_type Type, typename TCombiner, constraint::observable ...TOtherObservable>
+struct combine_latest_impl;
+
+template<constraint::decayed_type Type, typename SpecificObservable>
+struct member_overload<Type, SpecificObservable, combine_latest_tag>
+{
+
+    /**
+     * \brief Combines latest emissions from current observable and other observables when any of them emits.
+     *
```
victimsnino, line 38: This operator uses mutex to satisfy contract, so, it is better to add info about it (like in some other operators)

src/tests/benchmarks.cpp
```diff
@@ -391,6 +391,47 @@ TEST_CASE("with_latest_from")
...
+
+        rpp::observable::create<inner_source_type>([&](const auto& sub)
+            {
+                meter.measure([&]
+                {
+                    sub.on_next(inner_source);
+                });
+            })
```
victimsnino, line 429:
```suggestion
        rpp::observable::create<int>([&](const auto& sub)
            {
                meter.measure([&]
                {
                    sub.on_next(1);
                });
            })
```

src/tests/benchmarks.cpp
```diff
@@ -391,6 +391,47 @@ TEST_CASE("with_latest_from")
...
+{
+    BENCHMARK_ADVANCED("combine_latest construction from observable via dot + subscribe")(Catch::Benchmark::Chronometer meter)
+    {
+        const auto obs_1 = rpp::observable::create<int>([](const auto& subscriber)
+        {
+            subscriber.on_next(1);
+            subscriber.on_next(2);
+            subscriber.on_next(3);
+        });
+        const auto obs_2 = rpp::observable::create<int>([](const auto& subscriber)
```
victimsnino, line 404: you can reuse obs_1 as obs_2 there

victimsnino: + Same for RXCPP

src/tests/benchmarks.cpp
victimsnino: + Same for RXCPP

victimsnino (dismissed):

src/tests/test_combine_latest.cpp
```diff
@@ -0,0 +1,155 @@
...
+            .combine_latest(rpp::source::error<int>(std::make_exception_ptr(std::runtime_error{""})))
+            .subscribe(mock);
+
+        CHECK(mock.get_received_values().empty());
+        CHECK(mock.get_on_completed_count() == 0);
+        CHECK(mock.get_on_error_count() == 1);
+    }
+}
+
+SCENARIO("combine_latest regards unsubscription", "[combine_latest]")
```
victimsnino, line 122: These tests uselsess, because subscribe with unsubscribed observable do nothing [link](https://github.com/victimsnino/ReactivePlusPlus/blob/b87592fda89d76ac203ab6359d424dad9934fafa/src/rpp/rpp/observables/specific_observable.hpp#L63)

victimsnino (approved):

PR #220, opened 2022-08-20
Add "first" operator

tcw165:


src/tests/test_first.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+SCENARIO("first regards completed", "[first]")
+{
+    GIVEN("observable of ---|")
+    {
+        auto mock = mock_observer<int>{};
+        auto obs = rpp::source::empty<int>();
+        obs.first()
+            .subscribe(mock);
+        THEN("shall see ---|")
```
victimsnino, line 80: Shall see on_error with something like “No items”

tcw165: image

I feel that's the requirement for `Single` source.

victimsnino: Rxjava has next one

> The takeFirst operator behaves similarly to first, with the exception of how these operators behave wihen the source Observable emits no items that satisfy the predicate. In such a case, first will throw a NoSuchElementException while takeFirst will return an empty Observable (one that calls onCompleted but never calls onNext).

victimsnino: rxcpp also follows this one

victimsnino: Also it makes more sense. If I want to get first IF it appears, I can do take(1), but if I want to see first and get error in case of empty source, then first makes sense.

tcw165: That makes sense. I'll fix it.

src/rpp/rpp/operators.hpp
```diff
@@ -37,10 +37,11 @@
  * \ingroup operators
  */
 
-#include <rpp/operators/filter.hpp>
-#include <rpp/operators/take.hpp>
 #include <rpp/operators/distinct_until_changed.hpp>
+#include <rpp/operators/filter.hpp>
+#include <rpp/operators/first.hpp>
 #include <rpp/operators/skip.hpp>
+#include <rpp/operators/take.hpp>
```
tcw165, line 44: @victimsnino I also sort the lines in alphabetic order.

victimsnino:
I prefer to reuse `take_on_next` somehow.. Inherit from it or use it via composition, because on_next from `first` is 100% equal to on_next fro m`take_on_next`

tcw165:
> I prefer to reuse take_on_next somehow.. Inherit from it or use it via composition

I thought about that too. But then I did not follow that approach for these reasons:
1. Due to the reason we make sure ONE emission holds (zero emission is not allowed), we need to pollute `take_on_next{}`.
2. You have the tag dispatching for guarding the header import. If I remember it correctly, Kirk Shoop told me that the tag dispatching import allow the compiler to trim out the unused operators. From there, there might be a case where `take()` is used and `first()` is not. If we make `first()` reuse `take()` then we have a bigger `take()` that doesn't use all its ability.

tcw165:
Alright, I found the `take_on_next` isn't really bloated so I re-use it in my latest implementation.

tcw165:
@victimsnino I'm lost for the error message. Can you guide me?

victimsnino:
> @victimsnino I'm lost for the error message. Can you guide me?

Update branch and fix new compile errors. They appeared due to my recent change

src/rpp/rpp/operators/first.hpp
```diff
@@ -0,0 +1,75 @@
...
+           m_take_on_next(std::forward<decltype(value)>(value), subscriber);
+        }
+
+        void forward_on_completed(const constraint::subscriber auto& subscriber) const
+        {
+            if (!subscriber.is_subscribed()) return;
+
+            if (m_take_on_next.emission_count() != 0)
+            {
+                auto err = std::make_exception_ptr(std::runtime_error{"there should be one emission before completed"});
```
victimsnino, line 44: Can you create separate type of exception for this one, place it inside "fwd/first.hpp" and add comment inside documentation for "first" that it can \throws such type of exception?

src/rpp/rpp/operators/fwd/first.hpp
```diff
@@ -0,0 +1,74 @@
...
+template<constraint::decayed_type Type>
+struct first_impl;
+
+template<constraint::decayed_type Type, typename SpecificObservable>
+struct member_overload<Type, SpecificObservable, first_tag>
+{
+    /**
+     * \brief emit only the first item.
+     *
+     * \warning According to observable contract (https://reactivex.io/documentation/contract.html) emissions from any observable should be serialized, so, resulting observable uses mutex to satisfy this requirement
```
victimsnino, line 36: Not related to this one

src/rpp/rpp/operators/take.hpp
```diff
@@ -19,13 +19,22 @@ IMPLEMENTATION_FILE(take_tag);
...
 {
+
+template<constraint::decayed_type Type>
 class take_on_next
 {
 public:
-    take_on_next(size_t count) : m_count{ count } {}
+    explicit take_on_next(size_t count) : m_count{ count } {}
+
+    size_t emission_count() const
```
victimsnino, line 29:
```suggestion
    size_t get_emission_count_left() const
```

src/rpp/rpp/operators/take.hpp
```diff
@@ -19,13 +19,22 @@ IMPLEMENTATION_FILE(take_tag);
...
+    {
+        return m_count;
+    }
 
     void operator()(auto&& value, const constraint::subscriber auto& subscriber) const
     {
+        if (!subscriber.is_subscribed()) return;
```
victimsnino, line 36: You don't need this check at all,

src/rpp/rpp/operators/take.hpp
```diff
@@ -19,13 +19,22 @@ IMPLEMENTATION_FILE(take_tag);
 
 namespace rpp::details
 {
+
+template<constraint::decayed_type Type>
```
victimsnino, line 23: Why do we need it?

src/samples/doxygen/first.cpp
```diff
@@ -0,0 +1,20 @@
+#include <rpp/rpp.hpp>
+#include <iostream>
+
+/**
+ * \example first.cpp
+ **/
+int main()
+{
+    //! [first]
```
victimsnino, line 9: What about adding example with "empty" and exception?

src/tests/test_first.cpp
```diff
@@ -0,0 +1,122 @@
...
+        THEN("shall see -x")
+        {
+            CHECK(mock.get_received_values().empty());
+            CHECK(mock.get_on_completed_count() == 0);
+            CHECK(mock.get_on_error_count() == 1);
+        }
+    }
+}
+
+SCENARIO("first regards unsubscription", "[first]")
```
victimsnino, line 102: Not sure if it checks something related to first. Subscriber would not obtain any emissions if subscription is unsubscribed

src/rpp/rpp/operators/first.hpp
```diff
@@ -0,0 +1,75 @@
...
+    struct first_value_forwarder
+    {
+        void forward_on_next(auto&& value, const constraint::subscriber auto& subscriber) const
+        {
+           m_take_on_next(std::forward<decltype(value)>(value), subscriber);
+        }
+
+        void forward_on_completed(const constraint::subscriber auto& subscriber) const
+        {
+            if (!subscriber.is_subscribed()) return;
```
victimsnino, line 40: You don't need it

src/rpp/rpp/operators/take.hpp
tcw165: Oh why? It's for the case that this is subscribed to an unsubscribed subscriber.

src/tests/test_first.cpp
tcw165: Oh that's already taken care of in the subscriber and thus we don't need to reinforce that in lift's operator?

victimsnino: Yes, absolutely. There is no way send something to "unsubscribed" subscriber. Because
1) unsubscribed subscriber can't subscribe to new observable
2) unsubscribed subscriber can't get any on_next/on_error/on_completed due to extra checks


Only one way when we need to do it: inside "infinite-like" observables. For example, if we iterate over vector, then we don't need to continue iteration in case of subscriber unsubscribed

victimsnino: Moreover, if downstream and upstream subscribers use same subscription/child, then you can't even go to "operator's on_next" due to subscriber unsubcribed and it means that your "operator's subscriber" unsubscribed too and it can't accept new value

tcw165: > Moreover, if downstream and upstream subscribers use same subscription/child, then you can't even go to "operator's on_next" due to subscriber unsubcribed and it means that your "operator's subscriber" unsubscribed too and it can't accept new value

This is also a good topic to add it in my article collection.

tcw165: > Only one way when we need to do it: inside "infinite-like" observables.

Yup I got it. It makes more sense to add `is_subscribed()` check in the source.

src/rpp/rpp/operators/fwd/first.hpp
```diff
@@ -0,0 +1,78 @@
...
+struct first_tag;
+}
+
+namespace rpp::details
+{
+
+struct unmatched_emission_count_exception
+{
+    std::string what{"expect one emission before completed"};
+};
```
victimsnino, line 31: No, no, no. Any exceptions in C++ should be inherited from `std::runtime_error` in case of something wrong during runtime and `std::logic_error` if something unexpected happens. Also, the name of exception doesn't make sense. So, in your case it should be something like this
```suggestion
struct not_enough_emissions : std::runtime_error
{
    using std::runtime_error::runtime_error;
};
```
You can place it in something like `rpp/utils/exceptions.hpp` and use there. During throwing you can add details about exception something like "the first() operator requires at least one emission before complete"

src/tests/test_take.cpp
```diff
@@ -69,3 +69,25 @@ SCENARIO("take limits count of items", "[take]")
         }
     }
 }
+
+SCENARIO("take regards unsubscription", "[take]")
```
victimsnino, line 73: as discussed before - it doesn't make any sense

src/samples/doxygen/first.cpp
```diff
@@ -0,0 +1,31 @@
...
+        .first()
+        .subscribe(
+            [](const auto &v) { std::cout << "-" << v; },
+            [](const std::exception_ptr &error) {},
+            []() { std::cout << "-|" << std::endl; });
+    // Source: -1-2-3-4-5--|
+    // Output: -1-|
+    //! [first]
+
+    //! [first][empty]
```
victimsnino, line 20: Doxygen foramt to detect samples looks like `! [SOMETHING MEANINFUL]`. Range before  two such a blocks uses as example. So, change it there and below
```suggestion
    //! [first_empty]
```

src/rpp/rpp/operators/fwd/first.hpp
```diff
@@ -0,0 +1,78 @@
...
+         {
+             source observable   : +--1--2--3--|
+             operator "first"    : +--1-|
+         }
+     *
+     * \return new specific_observable with the first operator as most recent operator.
+     * \warning #include <rpp/operators/first.hpp>
+     *
+     * \par Example:
+     * \snippet first.cpp first
```
victimsnino, line 52:
```suggestion
     * \snippet first.cpp first
     * \snippet first.cpp first_empy
```

src/rpp/rpp/operators/fwd/first.hpp
tcw165: Oh side question, I found some articles saying we could throw any types.

victimsnino: Yeah, it is true, C++ provides ability to throw anything as exception. But also it it better to make exception in tree-like way for better handling and understanding. For example, as I’ve mentioned before, there are two types of error: logic and runtime. And in most cases if we want to “catch all exceptions” we catch runtime because logic means something absolutely unexpected and application can’t continue to work. Same can be propagated to another thing: we can create our subtype “RPP_exception” and inherit all our exceptions from this one. So, in case of we/user want to catch only RPP exceptions we can easily do it

victimsnino (approved):

PR #221, opened 2022-08-20
Clean up "take" for better readability

tcw165:
The `take_last()` operator is also needed for `last()` operator.

### DIFF
--- a/src/rpp/rpp/operators/fwd/take.hpp
+++ b/src/rpp/rpp/operators/fwd/take.hpp
@@ -46,13 +46,24 @@ struct member_overload<Type, SpecificObservable, take_tag>
     template<typename...Args>
     auto take(size_t count) const & requires is_header_included<take_tag, Args...>
     {
-        return static_cast<const SpecificObservable*>(this)->template lift<Type>(take_impl<Type>{count});
+        return cast_this()->template lift<Type>(take_impl<Type>{count});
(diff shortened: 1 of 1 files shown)


victimsnino:
I suggest to place it into separate header due to implementation is has another core logic comparing to take + RPP follows principle - 1 operator = 1 header

tcw165:
Yup I agree w/ you on that. I'll repurpose this PR to just a small clean-up.

src/rpp/rpp/operators/fwd/take.hpp
```diff
@@ -46,13 +46,24 @@ struct member_overload<Type, SpecificObservable, take_tag>
     template<typename...Args>
     auto take(size_t count) const & requires is_header_included<take_tag, Args...>
     {
-        return static_cast<const SpecificObservable*>(this)->template lift<Type>(take_impl<Type>{count});
+        return cast_this()->template lift<Type>(take_impl<Type>{count});
```
victimsnino, line 49: Then you can return const & instead of const * to use `.lift` instead of `-> template lift` :-)

tcw165: Hmm, this is not up-to-dated code here comparing to up-to-dated code in branch `tcw165:op/take`.

PR #235, opened 2022-08-27
Add "take_until" operator.

tcw165:


src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,132 @@
...
+//
+//    void decrement_notification_count(const auto& subscriber,
+//                                      const std::shared_ptr<first_state>& state) const
+//    {
+//        if (state->count > 0)
+//            --state->count;
+//
+//        if (state->count == 0)
+//            subscriber.on_completed();
+//    }
```
tcw165, line 94: @victimsnino 

> The TakeUntil subscribes and begins mirroring the source Observable. It also monitors a second Observable that you provide. If this second Observable emits an item or sends a termination notification, the Observable returned by TakeUntil stops mirroring the source Observable and terminates.

I'm not sure if `take_until()` will translate an error event to a completed event and hence I have this commented code.

You could check out the unit test and if the unit test is correct, I'll remove these lines.

victimsnino: Looks like your current implementation is correct (until observable should forward notification to subscriber)

src/tests/test_take_until.cpp
```diff
@@ -0,0 +1,94 @@
...
+            CHECK(mock.get_on_completed_count() == 1);
+            CHECK(mock.get_on_error_count() == 0);
+        }
+    }
+
+    GIVEN("observable of never pairs with trigger observable of -x")
+    {
+        auto mock = mock_observer<int>{};
+
+        rpp::source::never<int>()
```
victimsnino, line 83: Same as above

victimsnino (changes requested):

tcw165:
Rebase

src/rpp/rpp/operators/fwd/take_until.hpp
```diff
@@ -0,0 +1,81 @@
...
+     * \snippet take_until.cpp terminate
+     *
+     * \ingroup conditional_operators
+     * \see https://reactivex.io/documentation/operators/takeuntil.html
+     */
+    template<constraint::observable TTriggerObservable>
+    auto take_until(TTriggerObservable&& until_observable) const& requires is_header_included<take_until_tag, TTriggerObservable>
+    {
+        return cast_this()->template lift<Type>(
+            take_until_impl<Type, TTriggerObservable>{std::forward<TTriggerObservable>(until_observable)});
```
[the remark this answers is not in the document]
tcw165: Thanks, this is important and I keep forgetting about this.

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * \brief "combine_latest" operator (an OperatorFn used by "lift").
+ */
+template<constraint::decayed_type Type, constraint::observable TTriggerObservable>
+struct take_until_impl
+{
+    using TriggerType = utils::extract_observable_type_t<TTriggerObservable>;
+
+    const TTriggerObservable m_until_observable;
```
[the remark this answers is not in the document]
tcw165: q: You mean implicit `m_until_observable =` using move assignment operator in constructor?

q: Why is it bad for disabling "move" constructor here? Is it safer that we disable "move" constructor?

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,132 @@
...
+    {
+        auto state = std::make_shared<first_state>();
+
+        // Subscribe to trigger observable
+        auto child_subscription = subscriber.get_subscription().make_child();
+        m_until_observable.subscribe(
+            create_subscriber_with_state<TriggerType>(std::move(child_subscription),
+                                                      take_until_throttler<TriggerType>{},
+                                                      utils::forwarding_on_error{},
+                                                      utils::forwarding_on_completed{},
```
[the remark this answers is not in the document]
tcw165: Let me rephrase this case:

```
source observable        : +-1--2--3--|
until observable         : +-s--s-----|
operator "take_until"    : +-1--------|
```

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,132 @@
...
+struct take_until_impl
+{
+    using TriggerType = utils::extract_observable_type_t<TTriggerObservable>;
+
+    const TTriggerObservable m_until_observable;
+
+    template<constraint::subscriber_of_type<Type> TSub>
+    auto operator()(TSub&& subscriber) const
+    {
+        auto state = std::make_shared<first_state>();
```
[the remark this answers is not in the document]
tcw165: Oh it's for sharing in between two subscribers:

```
auto child_subscription = subscriber.get_subscription().make_child();
m_until_observable.subscribe(
    create_subscriber_with_state<TriggerType>(...
                                              state));

auto subscription = subscriber.get_subscription();
return create_subscriber_with_state<Type>(...
                                          std::move(state));
```

src/rpp/rpp/operators/take_until.hpp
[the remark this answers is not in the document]
victimsnino: No, i mean remove `const` qualifier.

>  Is it safer that we disable "move" constructor?

In this case- no. It is state of `_impl` - so, it is state of observable, not subscriber. As a result, any observable with `take_until` on board would not be able to be moved, but it is bad. For example
```cpp
  obs.map(...)
     .take_untile(...) // starting from this one no any moves would be possible, only copies
     .filter(...)
     .take(...)
```

src/rpp/rpp/operators/take_until.hpp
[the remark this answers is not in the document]
victimsnino: yes

src/rpp/rpp/operators/take_until.hpp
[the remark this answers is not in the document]
victimsnino: You didn't catch. Why `first_state` not something like `take_last_state` ?

src/rpp/rpp/operators/fwd/take_until.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * \brief Discard any items emitted by an Observable after a second Observable emits an item or terminates.
+     * \warning The take_until subscribes and begins mirroring the source Observable. It also monitors a second Observable that you provide. If this second Observable emits an item or sends a on_error/on_completed notification, the Observable returned by take_until stops mirroring the source Observable and terminates.
+     *
+     * \marble take_until
+       {
+           source observable        : +-1--2--3--|
+           until observable         : +--s--s----|
+           operator "take_until"    : +-1--------|
```
victimsnino, line 41: Actually, it would be something like this
```suggestion
           operator "take_until"    : +-1|
```

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,108 @@
...
+//          https://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/victimsnino/ReactivePlusPlus
+//
+
+#pragma once
+
+#include <rpp/defs.hpp>
+#include <rpp/operators/fwd/take_until.hpp>
+#include <rpp/operators/first.hpp>
```
victimsnino, line 16: Looks like we don't need it

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,108 @@
...
+#include <rpp/sources/create.hpp>
```
victimsnino, line 17: Looks like we don't need it too

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,108 @@
...
+    void operator()(auto&& value,
+                    const auto& subscriber,
+                    const std::shared_ptr<take_until_state>& state) const
+    {
+        std::lock_guard lock{state->mutex};
+
+        if (!state->is_stopped)
+            subscriber.on_next(std::forward<decltype(value)>(value));
+        else
+            subscriber.on_completed();
```
victimsnino, line 51: Looks like this line useless. If `is_stopped` true, then we've send `on_completed`/`on_error` from another observable

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,108 @@
...
+    {
+        auto state = std::make_shared<take_until_state>();
+
+        // Subscribe to trigger observable
+        auto child_subscription = subscriber.get_subscription().make_child();
+        m_until_observable.subscribe(
+            create_subscriber_with_state<TriggerType>(std::move(child_subscription),
+                                                      take_until_throttler<TriggerType>{},
+                                                      utils::forwarding_on_error{},
+                                                      utils::forwarding_on_completed{},
```
victimsnino, line 94: looks like these functions also should be under lock with check over "is_stopped" (same about another subscriber)

src/tests/test_take_until.cpp
```diff
@@ -0,0 +1,107 @@
...
+SCENARIO("take_until handles race condition", "[take_until]")
+{
+    GIVEN("trigger value is emitted when source on_next() is still emitting.")
+    {
+        auto mock = mock_observer<size_t>{};
+
+        rpp::source::interval(std::chrono::seconds{1}, rpp::schedulers::trampoline{})
+            .take_until(rpp::source::interval(std::chrono::seconds{1}, rpp::schedulers::trampoline{}))
+            .subscribe(mock);
+    }
```
victimsnino, line 106: There is no any race-condition =) Because everything happens at the same thread =)

victimsnino (commented):
Overall looks much better

src/tests/test_take_until.cpp
tcw165: Oh my bad, I accidentally commit this and I know this doesn't simulate race-condition at all.

src/rpp/rpp/operators/take_until.hpp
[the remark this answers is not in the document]
tcw165: I see. You were asking why `first_state` rather than `take_last_state`.

src/rpp/rpp/operators/take_until.hpp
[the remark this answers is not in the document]
tcw165: How to simulate the race condition BTW? Can we use scheduler with timestamp to do so?

src/rpp/rpp/operators/take_until.hpp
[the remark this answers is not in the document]
tcw165: Oh this is very interesting, I didn't know the observable may be copied in a chain. I thought the observable is always moved in order to save CPU time and memory. 

How do I know what operator copies the observable and what moves the observable?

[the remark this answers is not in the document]
tcw165: Also, if `take_until` moves the observable during the subscribing, the subsequent subscribing (could be triggered by `retry` family of operators) would get a segmentation fault?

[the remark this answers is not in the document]
victimsnino: No, you didn't get my point.

When we call something like `my_obs.map(...)` we actually call `map` function with `const &` qualifier. For example
```cpp
template<std::invocable<Type> Callable>
    auto map(Callable&& callable) const & requires is_header_included<map_tag, Callable>
    {
        return static_cast<const SpecificObservable*>(this)->template lift<utils::decayed_invoke_result_t<Callable, Type>>(map_impl<Type, std::decay_t<Callable>>{std::forward<Callable>(callable)});
    }

```

What do we do there? Actually, we call `lift` from lvalue observable. Due to we construct new observable, we need to copy current to extend chain. So, observable is copied, because we need to keep both (`my_obs` and new one) as valid. Do you get it?

Then, we also can call same `map` as rvalue in both cases
1) `std::move(my_obs).map(...)` - manually move it
2) `my_obs.take(1).map(...)` - results of `take` is new observable, but currently is `prvalue`

In this case another overloading would be called:

```cpp
template<std::invocable<Type> Callable>
    auto map(Callable&& callable) && requires is_header_included<map_tag, Callable>
    {
        return std::move(*static_cast<SpecificObservable*>(this)).template lift<utils::decayed_invoke_result_t<Callable, Type>>(map_impl<Type, std::decay_t<Callable>>{std::forward<Callable>(callable)});
    }
```

It is rvalue overloading, so, we call lift as rvalue (via move). Same as above: to construct new observable we need have "old" observable to extend it. But due to we call it as rvalue we can "take" observable as is without copy. So, we just move current observable.

Observable is just "on subscribe" function. If "on subscribe" is not movable, then any move would be actually copy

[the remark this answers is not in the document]
victimsnino: So, these copies/moves happen at the stage of construction of the chain, not subscribing

src/rpp/rpp/operators/take_until.hpp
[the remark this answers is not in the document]
victimsnino: You need to add multi-threading. Something like this:

```cpp
bool on_completed_called{};
rpp::source::interval(std::chrono::seconds{1}, rpp::schedulers::trampoline{})
            .take_until(rpp::source::interval(std::chrono::seconds{2}, rpp::schedulers::new_thread{}))
            .subscribe([](const auto&)
            {
                CHECK(!on_completed_called);
                std::this_thread::sleep_for(std::chrono::seconds(3));
                CHECK(!on_completed_called);
            },
            [](){on_completed_called = true;});
CHECK(on_completed_called);
```

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Functor (type-erasure) of "take_until" for on_completed operator.
+     */
+    void operator()(const auto& subscriber,
+                    const std::shared_ptr<take_until_state>& state) const
+    {
+        std::lock_guard lock{state->mutex};
+
+        subscriber.on_completed();
+    }
```
victimsnino, line 74: Split to 3 functors please

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Functor (type-erasure) of throttler (trigger observable) for on_completed operator.
+     */
+    void operator()(const auto& subscriber,
+                    const std::shared_ptr<take_until_state>& state) const
+    {
+        std::lock_guard lock{state->mutex};
+
+        subscriber.on_completed();
+    }
```
victimsnino, line 117: is_stopped there also should be set to `true`

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,159 @@
...
+        std::lock_guard lock{state->mutex};
+
+        subscriber.on_error(err);
+    }
+};
+
+/**
+ * Functor (type-erasure) of throttler (trigger observable) for on_completed operator.
+ */
+struct take_until_throttler_on_completed
```
victimsnino, line 112:
```suggestion
using take_until_throttler_on_completed = take_until_on_completed;
```

src/rpp/rpp/operators/take_until.hpp
```diff
@@ -0,0 +1,159 @@
...
+
+        state->is_stopped = true;
+        subscriber.on_completed();
+    }
+};
+
+/**
+ * Functor (type-erasure) of throttler (trigger observable) for on_error operator.
+ */
+struct take_until_throttler_on_error {
```
victimsnino, line 98:
```suggestion
using take_until_throttler_on_error = take_until_on_error;
```

src/tests/test_take_until.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+        THEN("should see -| because take_until is subscribed first and it also mirrors the trigger observable 's completed event")
+        {
+            CHECK(mock.get_received_values().empty());
+            CHECK(mock.get_on_completed_count() == 1);
+            CHECK(mock.get_on_error_count() == 0);
+        }
+    }
+
+    GIVEN("observable of -1-| pairs with trigger observable of -x")
```
victimsnino, line 82: Add also test "with trigget observable of -|"

src/tests/test_take_until.cpp
```diff
@@ -0,0 +1,121 @@
...
+                           CHECK(!on_completed_called);
+                       } /* on_next */,
+                       {} /* on_error */,
+                       [&]()
+                       {
+                           on_completed_called = true;
+                       } /* on_completed */);
+
+        CHECK(on_completed_called);
+    }
```
victimsnino, line 120:
```suggestion
        {
        std::atomic_bool on_completed_called{false};
        auto             subject = rpp::subjects::publish_subject<int>{};

        WHEN("subscribe on it")
        {
            THEN("on_error can't interleave with on_next")
            {
                rpp::source::just(1, 1, 1)
                        .take_until(subject.get_observable())
                        .as_blocking()
                        .subscribe([&](auto&&)
                                   {
                                       CHECK(!on_error_called);
                                       std::thread{[&]
                                       {
                                           subject.get_subscriber().on_error(std::exception_ptr{});
                                       }}.detach();
                                       std::this_thread::sleep_for(std::chrono::seconds{1});
                                       CHECK(!on_error_called);
                                   },
                                   [&]() { on_completed_called= true; });

                CHECK(on_error_called);
            }
        }
    }
```

src/rpp/rpp/operators/take_until.hpp
tcw165: Smart!

victimsnino (dismissed):

tcw165:
@victimsnino The "early-unsubscribing" fixes the race condition among `on_next`, `on_error` and `on_completed` dispatching.

image

victimsnino (approved):

PR #253, opened 2022-09-05
Reinforce race-condition for "combine_latest" operator.

tcw165:


tcw165:
With unit test, we could reproduce:

```
/tmp/rpp/src/tests/test_combine_latest.cpp:146: FAILED:
  CHECK( !on_error_called )
with expansion:
  false

===============================================================================
test cases:  4 |  3 passed | 1 failed
assertions: 20 | 19 passed | 1 failed
```

PR #254, opened 2022-09-06
nit: Give better name for on_subscribe functor.

tcw165:


src/rpp/rpp/observables/specific_observable.hpp
```diff
@@ -104,7 +104,10 @@ class specific_observable : public interface_observable<Type, specific_observabl
     }
 
 private:
-    RPP_NO_UNIQUE_ADDRESS OnSubscribeFn m_state;
+    /**
+     * \brief The on_subscribe functor (type-erasure), which has the operator()(const auto& subscriber) overload function.
```
victimsnino, line 108: There is no any type-erasure :) we use some specific type,not something like std::function

victimsnino (approved):

Issue #266, opened 2022-09-08
Add "delay" operator


victimsnino:
@wuyuanyi135 requested "delay" operator

tcw165:
I could own it.

PR #269, opened 2022-09-09
Add "delay" operator

tcw165:
For https://github.com/victimsnino/ReactivePlusPlus/issues/266

### DIFF
--- a/src/examples/doxygen/delay.cpp
+++ b/src/examples/doxygen/delay.cpp
@@ -1,0 +1,24 @@
+#include <rpp/rpp.hpp>
+
+#include <chrono>
+#include <iostream>
+
+/**
+ * \example delay.cpp
+ **/
+int main()
+{
+    //! [delay]
+    std::cout << std::this_thread::get_id() << std::endl;
+    rpp::source::just(1, 2, 3)
+            .delay(std::chrono::seconds{1}, rpp::schedulers::new_thread{})
+            .as_blocking()
+            .subscribe([](int v) { std::cout << "[" << std::this_thread::get_id() << "] : " << v << std::endl; });
+    // Template for output:
+    // TH1
+    // [TH2]: 1
+    // [TH2]: 2
+    // [TH2]: 3
+    //! [delay]
+    return 0;
+}
--- a/src/examples/doxygen/delay.cpp
+++ b/src/examples/doxygen/delay.cpp
@@ -1,6 +1,7 @@
 #include <rpp/rpp.hpp>
 
 #include <chrono>
+#include <ctime>
 #include <iostream>
 
 /**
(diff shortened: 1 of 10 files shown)


src/examples/doxygen/delay.cpp
```diff
@@ -0,0 +1,24 @@
...
+ * \example delay.cpp
+ **/
+int main()
+{
+    //! [delay]
+    std::cout << std::this_thread::get_id() << std::endl;
+    rpp::source::just(1, 2, 3)
+            .delay(std::chrono::seconds{1}, rpp::schedulers::new_thread{})
+            .as_blocking()
+            .subscribe([](int v) { std::cout << "[" << std::this_thread::get_id() << "] : " << v << std::endl; });
```
victimsnino, line 16: I think it is also good idea to show diff in time. For example "time since subscription" because `delay` it is more about delaying, not using scheduler

src/rpp/rpp/operators/fwd/delay.hpp
```diff
@@ -0,0 +1,83 @@
...
+        return cast_this()->template lift<Type>(
+            delay_impl<Type, std::decay_t<TScheduler>>{std::forward<TScheduler>(scheduler),
+                                                       std::chrono::duration_cast<rpp::schedulers::duration>(delay_duration)});
+    }
+
+    template<schedulers::constraint::scheduler TScheduler>
+    auto delay(auto&& delay_duration,
+               TScheduler&& scheduler) && requires is_header_included<delay_tag, TScheduler>
+    {
+        return cast_this()->template lift<Type>(
```
victimsnino, line 66: move_this, not cast_this

src/rpp/rpp/operators/fwd/delay.hpp
```diff
@@ -0,0 +1,83 @@
...
+     * \brief Shift the emissions from an Observable forward in time by a particular amount.
+     * \details The delay operator modifies its source Observable by pausing for a particular increment of time (that you specify) before emitting each of the source Observable’s items. This has the effect of shifting the entire sequence of items emitted by the Observable forward in time by that specified increment.
+     *
+     * \marble delay
+       {
+           source observable        : +-1-2-3-|
+           operator "delay: --"     : +---1-2-3-|
+       }
+     *
+     * \param delay_duration is the delay duration for emitting items. Delay duration should be able to cast to rpp::schedulers::duration.
```
victimsnino, line 43: add \param scheduler

src/rpp/rpp/operators/delay.hpp
```diff
@@ -0,0 +1,111 @@
...
+        auto dynamic_subscriber = std::forward<TSub>(subscriber).as_dynamic();
+        auto subscription = dynamic_subscriber.get_subscription().make_child();
+
+        auto worker = scheduler.create_worker(dynamic_subscriber.get_subscription());
+
+        return create_subscriber_with_state<Type>(std::move(subscription),
+                                                  delay_on_next{delay},
+                                                  delay_on_error{delay},
+                                                  delay_on_completed{delay},
+                                                  dynamic_subscriber,
```
victimsnino, line 106: (minor) also can be moved =)

src/tests/test_delay.cpp
```diff
@@ -0,0 +1,84 @@
...
+        const auto now = rpp::schedulers::clock_type::now();
+
+        rpp::source::just(1)
+            .delay(delay_duration, rpp::schedulers::trampoline{})
+            .as_blocking()
+            .subscribe(mock);
+
+        THEN("should see -1-| after the delay")
+        {
+            CHECK(rpp::schedulers::clock_type::now() >= now + std::chrono::duration_cast<rpp::schedulers::duration>(delay_duration));
```
victimsnino, line 37: To be honest, it is better to do this check INSIDE `on_next`. It would be more correct and understandable

victimsnino (changes requested):
Good!

victimsnino (approved):

PR #276, opened 2022-09-17
Petri net executor example

thorstink:
As promised, an, imho fun and simplified example of implementing Petri net execution using reactive streams. 

Note that it could be a bit fancier by executing the transitions in a different context - that way new transitions would get queued while the active transition is still running. Once RPP has something similar to RxCpp's `observe_on_event_loop` it would also be straightforward to actually execute transitions parallel using Rx-patterns. 

Let me know if you think this example adds value to your project and if there's something you'd like to see clarified/improved.


victimsnino:
Yeah, looks cool! Thank you!
Could you address comments from CI to pass it? Then i would merge it

thorstink:
The ci output  https://github.com/victimsnino/ReactivePlusPlus/actions/runs/3074503201/jobs/4967539918 and https://github.com/victimsnino/ReactivePlusPlus/actions/runs/3074503201/jobs/4967539973 does not give me a clue what to look at (or I am missing it). Do you see it?

Thomas

victimsnino:
You need to unfold Build with Cmake inside this one

error:
```
Error: /home/runner/work/ReactivePlusPlus/ReactivePlusPlus/src/examples/petri/petri.cpp:84:26: error: local variable 'marking' will be copied despite being returned by name [-Werror,-Wreturn-std-move]
[363](https://github.com/victimsnino/ReactivePlusPlus/actions/runs/3074503201/jobs/4967539973#step:5:366)
                    return marking;
[364](https://github.com/victimsnino/ReactivePlusPlus/actions/runs/3074503201/jobs/4967539973#step:5:367)
                           ^~~~~~~
```

thorstink:
I was unable to easily reproduce this, clang-14 didn't give me this error (with the same flags as in the ci-matrix).. However, I think this edit should fix it.

victimsnino (approved):

thorstink:
A very cool! Will check it out. 

Thomas

PR #279, opened 2022-10-10
Add "on_error_resume_next" operator.

tcw165:


tcw165:
@victimsnino The office ReactiveX doc names the operator `catch()`, and RxCpp names it `on_error_resume_next()`. Shall we just use `catch()` for consistent naming with the doc?

victimsnino:
> @victimsnino The office ReactiveX doc names the operator `catch()`, and RxCpp names it `on_error_resume_next()`. Shall we just use `catch()` for consistent naming with the doc?

I offer to provide aliases for both ways to follow ReactiveX and RxCpp and the same time

tcw165:
I see. That makes sense. Let's keep using `on_error_resume_next()` though.

src/rpp/rpp/operators/on_error_resume_next.hpp
```diff
@@ -0,0 +1,85 @@
...
+{
+    RPP_NO_UNIQUE_ADDRESS ResumeCallable m_resume_callable;
+
+    explicit on_error_resume_next_impl(ResumeCallable&& callable) : m_resume_callable(std::move(callable)) {};
+    explicit on_error_resume_next_impl() = delete;
+
+    template<constraint::subscriber_of_type<Type> TSub>
+    auto operator()(TSub&& downstream_subscriber) const
+    {
+        auto state = std::make_shared<on_error_resume_state<ResumeCallable>>(downstream_subscriber.get_subscription(), m_resume_callable);
```
victimsnino, line 73: you don't need shared_ptr there. Subscription is just "shared". Callable should be immutable, so, also can be easily copied. So, pass them as is. I think, it is main reason why our implementation is a bit slower than rxcpp in some cases

src/rpp/rpp/operators/on_error_resume_next.hpp
```diff
@@ -0,0 +1,85 @@
...
+    void operator()(const std::exception_ptr& err,
+                    const auto& subscriber,
+                    const std::shared_ptr<on_error_resume_state<ResumeCallable>>& state) const
+    {
+        using Type = rpp::utils::extract_subscriber_type_t<decltype(subscriber)>;
+
+        auto new_observable = state->m_resume_callable(err);
+
+        // Subscribe to next_observable
+        new_observable.subscribe(create_subscriber_with_state<Type>(state->m_subscription,
```
victimsnino, line 51: subscriber.get_subscription() ?

src/rpp/rpp/operators/fwd/on_error_resume_next.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+template<constraint::decayed_type Type, rpp::details::resume_callable ResumeCallable>
+struct on_error_resume_next_impl;
+
+template<constraint::decayed_type Type, typename SpecificObservable>
+struct member_overload<Type, SpecificObservable, on_error_resume_next_tag>
+{
+
+    /**
+     * \brief Recover from an onError notification by continuing the sequence without error.
```
victimsnino, line 40:
```suggestion
     * \brief Recover from an on_error notification by continuing the sequence without error.
```

src/rpp/rpp/operators/on_error_resume_next.hpp
```diff
@@ -0,0 +1,85 @@
...
+template<rpp::details::resume_callable ResumeCallable>
+struct on_error_resume_state
+{
+    rpp::composite_subscription m_subscription;
+    ResumeCallable m_resume_callable;
+
+    explicit on_error_resume_state(rpp::composite_subscription subscription,
+                                   const ResumeCallable& callable) : m_subscription(std::move(subscription))
+                                                                   , m_resume_callable(callable) {};
+    explicit on_error_resume_state() = delete;
```
victimsnino, line 33: default constructor automatically "deleted" in case of any manually specified constructor

tcw165: Oh interestingly. Where is that statement stated? In https://en.cppreference.com/w/cpp/language/default_constructor ?

victimsnino: https://eel.is/c++draft/class.ctor#class.default.ctor-1.sentence-2

> If there is no user-declared constructor for class X, a non-explicit constructor having no parameters is implicitly declared as defaulted

```cpp
#include <optional>
#include <iostream>

struct NoCtr
{
    int v{};
};

struct Ctr
{
    Ctr(int v){}
    int v{};
};

int main()
{
    std::cout << "NoCtr " << std::is_default_constructible<NoCtr>() << std::endl;
    std::cout << "Ctr " << std::is_default_constructible<Ctr>() << std::endl;

    return 0;
}
```

```console
NoCtr 1
Ctr 0
```
[godbolt](https://godbolt.org/z/T5h6TEK7K)

victimsnino: > Oh interestingly. Where is that statement stated? In https://en.cppreference.com/w/cpp/language/default_constructor ?


Your link also has something similar
> If no user-declared constructors of any kind are provided for a class type (struct, class, or union), the compiler will always declare a default constructor as an inline public member of its class.

> If some user-declared constructors are present, the user may still force the automatic generation of a default constructor by the compiler that would be implicitly-declared otherwise with the keyword default.

victimsnino: So, it works in vice versa. Not deleted when has some user-defined ctor, but generated if no any user-defined

src/rpp/rpp/operators/fwd/on_error_resume_next.hpp
```diff
@@ -0,0 +1,84 @@
...
+     * \par Examples
+     * \snippet on_error_resume_next.cpp on_error_resume_next
+     *
+     * \ingroup error_handling_operators
+     * \see https://reactivex.io/documentation/operators/on_error_resume_next.html
+     */
+    template<rpp::details::resume_callable ResumeCallable>
+    auto on_error_resume_next(ResumeCallable&& resume_callable) const& requires is_header_included<on_error_resume_next_tag, ResumeCallable>
+    {
+        return cast_this()->template lift<Type>(on_error_resume_next_impl<Type, ResumeCallable>{std::forward<ResumeCallable>(resume_callable)});
```
victimsnino, line 63:
```suggestion
        return cast_this()->template lift<Type>(on_error_resume_next_impl<Type, std::decay_t<ResumeCallable>>{std::forward<ResumeCallable>(resume_callable)});
```

src/rpp/rpp/operators/fwd/on_error_resume_next.hpp
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    template<rpp::details::resume_callable ResumeCallable>
+    auto on_error_resume_next(ResumeCallable&& resume_callable) && requires is_header_included<on_error_resume_next_tag, ResumeCallable>
+    {
+        return move_this().template lift<Type>(on_error_resume_next_impl<Type, ResumeCallable>{std::forward<ResumeCallable>(resume_callable)});
```
victimsnino, line 69:
```suggestion
        return move_this().template lift<Type>(on_error_resume_next_impl<Type, std::decay_t<ResumeCallable>>{std::forward<ResumeCallable>(resume_callable)});
```

src/rpp/rpp/operators/on_error_resume_next.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+        auto new_observable = resume_callable(err);
+
+        // Subscribe to next_observable
+        auto subscription = subscriber.get_subscription();
+        new_observable.subscribe(create_subscriber_with_state<Type>(std::move(subscription),
+                                                                    rpp::utils::forwarding_on_next{},
+                                                                    rpp::utils::forwarding_on_error{},
+                                                                    rpp::utils::forwarding_on_completed{},
+                                                                    subscriber));
```
victimsnino, line 44: Amount of copies/moves can be reduced a bit
```suggestion
        // Subscribe to next_observable
        resume_callable(err).subscribe(create_subscriber_with_state<Type>(subscriber.get_subscription(),
                                                                    rpp::utils::forwarding_on_next{},
                                                                    rpp::utils::forwarding_on_error{},
                                                                    rpp::utils::forwarding_on_completed{},
                                                                    subscriber));
```

src/rpp/rpp/operators/on_error_resume_next.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * \brief Functor of OperatorFn for "on_error_resume_next" operator (used by "lift").
+ */
+template<constraint::decayed_type Type, rpp::details::resume_callable ResumeCallable>
+struct on_error_resume_next_impl
+{
+    RPP_NO_UNIQUE_ADDRESS ResumeCallable m_resume_callable;
+
+    explicit on_error_resume_next_impl(ResumeCallable&& callable) : m_resume_callable(std::move(callable)) {};
```
victimsnino, line 56: There is no guarantee, that callable is moved. It can be copied =) It is compilable due to missed decay_t in declaration

src/tests/test_on_error_resume_next.cpp
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+SCENARIO("on_error_resume_next forwards on_next and on_completed", "[on_error_resume_next]")
+{
+    GIVEN("observable of -1-2-3-|")
+    {
+        auto mock = mock_observer<int>{};
+
+        rpp::source::just(1, 2, 3)
```
victimsnino, line 85: I think, it also nice to have mixed observable: it emits, lets say, 1,2 then error and then we see values from "resume". Something like 
```cpp
rpp::sources::just(just(1), just(2), error(..)).concat()
```

victimsnino (changes requested):

src/rpp/rpp/operators/on_error_resume_next.hpp
tcw165: Discussed offline, https://godbolt.org/z/e9h5c9v9d

src/rpp/rpp/operators/on_error_resume_next.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * \brief Functor of OperatorFn for "on_error_resume_next" operator (used by "lift").
+ */
+template<constraint::decayed_type Type, rpp::details::resume_callable ResumeCallable>
+struct on_error_resume_next_impl
+{
+    RPP_NO_UNIQUE_ADDRESS ResumeCallable m_resume_callable;
+
+    explicit on_error_resume_next_impl(ResumeCallable&& callable) : m_resume_callable(std::forward<ResumeCallable>(callable)) {};
```
victimsnino, line 53:
```suggestion
    explicit on_error_resume_next_impl(const ResumeCallable& callable) : m_resume_callable(callable) {};
    explicit on_error_resume_next_impl(ResumeCallable&& callable) : m_resume_callable(std::move(callable)) {};
```

PR #280, opened 2022-10-11
Add "sample_with_time" operator

### LINKED ISSUE #265: Add "sample_with_time" operator
"nebu" from discord requested implementation of `sample` and `sample_with_time` operators

victimsnino:
Partly resolves #265

### DIFF
--- a/docs/Implementation Status.md
+++ b/docs/Implementation Status.md
@@ -71,6 +71,8 @@
 - [ ] ignore_elements
 - [x] last
 - [ ] sample
+  - [ ] sample (observable)
+  - [x] sample_with_time
 - [x] skip
 - [ ] skip_last
 - [x] take_last
--- a/src/examples/doxygen/sample.cpp
+++ b/src/examples/doxygen/sample.cpp
@@ -1,0 +1,17 @@
+#include <rpp/rpp.hpp>
+#include <iostream>
+
+/**
+ * \example sample.cpp
+ **/
+int main()
+{
+    //! [sample_with_time]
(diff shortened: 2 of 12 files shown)


thorstink:
Awesome, thanks :-)

victimsnino:
@thorstink, waiting for the feedback ;-)
BTW, do you still need sample with observable logic?

thorstink:
Ah, I think I am missing something. What feedback? :-)

sample_with_time is something I would still like a lot. What I mean is
something similar to RxCpp's sample_with_time:
https://github.com/ReactiveX/RxCpp/blob/main/Rx/v2/test/operators/sample.cpp

Something that rxcpp does not have, but what I think would be nice (I've
noticed I wanted it sometimes), is a simple downsample operator. Pass one
message, skip three, pass one, skip three etc.

Or did you mean something else?

victimsnino:
@thorstink ,

> Ah, I think I am missing something. What feedback? :-)

Feedback about usage =) As far as I remember you requested this operator =)

> sample_with_time is something I would still like a lot. What I mean is
something similar to RxCpp's sample_with_time:
https://github.com/ReactiveX/RxCpp/blob/main/Rx/v2/test/operators/sample.cpp

Yeah, rpp has same after this PR =)

> Something that rxcpp does not have, but what I think would be nice (I've
noticed I wanted it sometimes), is a simple downsample operator. Pass one
message, skip three, pass one, skip three etc.

Does it something like "take N elements each NNth element"?

PR #374, opened 2023-06-10
Implement flat_map operator

CorentinBT:
Should I also implement a variant taking a `ResultSelector` like in rxcpp?

### DIFF
--- a/src/benchmarks/benchmarks.cpp
+++ b/src/benchmarks/benchmarks.cpp
@@ -302,14 +302,14 @@ int main(int argc, char* argv[]) // NOLINT
             TEST_RPP([&]()
             {
                 rpp::source::create<int>([](const auto& obs){ obs.on_next(1); })
-                    | rpp::operators::flat_map([](int v) { return rpp::source::just(v * 2); })
+                    | rpp::operators::flat_map([](int v) {  return rpp::source::create<int>([v](const auto& obs){ obs.on_next(v * 2); }); })
                     | rpp::operators::subscribe([](int v){ ankerl::nanobench::doNotOptimizeAway(v); });
             });
 
             TEST_RXCPP([&]()
             {
                 rxcpp::observable<>::create<int>([](const auto& obs){obs.on_next(1);})
-                    | rxcpp::operators::flat_map([](int v) { return rxcpp::observable<>::just(v * 2); })
+                    | rxcpp::operators::flat_map([](int v) { return rxcpp::observable<>::create<int>([v](const auto& obs){ obs.on_next(v * 2); }); })
                     | rxcpp::operators::subscribe<int>([](int v){ ankerl::nanobench::doNotOptimizeAway(v); });
             });
         }
--- a/src/benchmarks/benchmarks.cpp
+++ b/src/benchmarks/benchmarks.cpp
@@ -297,6 +297,22 @@ int main(int argc, char* argv[]) // NOLINT
                     | rxcpp::operators::subscribe<int>([](int v){ ankerl::nanobench::doNotOptimizeAway(v); });
             });
         }
+        SECTION("create+flat_map(just(v*2))+subscribe")
+        {
+            TEST_RPP([&]()
+            {
+                rpp::source::create<int>([](const auto& obs){ obs.on_next(1); })
+                    | rpp::operators::flat_map([](int v) { return rpp::source::just(v * 2); })
+                    | rpp::operators::subscribe([](int v){ ankerl::nanobench::doNotOptimizeAway(v); });
+            });
+
+            TEST_RXCPP([&]()
+            {
+                rxcpp::observable<>::create<int>([](const auto& obs){obs.on_next(1);})
+                    | rxcpp::operators::flat_map([](int v) { return rxcpp::observable<>::just(v * 2); })
+                    | rxcpp::operators::subscribe<int>([](int v){ ankerl::nanobench::doNotOptimizeAway(v); });
+            });
+        }
     };
 
     BENCHMARK("Filtering Operators")
--- a/src/rpp/rpp/operators/flat_map.hpp
+++ b/src/rpp/rpp/operators/flat_map.hpp
@@ -22,8 +22,9 @@ struct flat_map_t
 {
     RPP_NO_UNIQUE_ADDRESS Fn m_fn;
 
-    template<rpp::constraint::observable TObservable, typename ValueType = rpp::utils::extract_observable_type_t<TObservable>>
-        requires (std::invocable<Fn, ValueType> && rpp::constraint::observable<std::invoke_result_t<Fn, ValueType>>)
+    template<rpp::constraint::observable TObservable>
+        requires (std::invocable<Fn, rpp::utils::extract_observable_type_t<TObservable>> 
+                    && rpp::constraint::observable<std::invoke_result_t<Fn, rpp::utils::extract_observable_type_t<TObservable>>>)
     auto operator()(TObservable&& observable) const &
     {
         return std::forward<TObservable>(observable)
@@ -32,8 +33,9 @@ struct flat_map_t
                 
     }
 
-    template<rpp::constraint::observable TObservable, typename ValueType = rpp::utils::extract_observable_type_t<TObservable>>
-        requires (std::invocable<Fn, ValueType> && rpp::constraint::observable<std::invoke_result_t<Fn, ValueType>>)
+    template<rpp::constraint::observable TObservable>
+        requires (std::invocable<Fn, rpp::utils::extract_observable_type_t<TObservable>> 
+                    && rpp::constraint::observable<std::invoke_result_t<Fn, rpp::utils::extract_observable_type_t<TObservable>>>)
     auto operator()(TObservable&& observable) &&
     {
         return std::forward<TObservable>(observable)
--- a/src/rpp/rpp/operators/flat_map.hpp
+++ b/src/rpp/rpp/operators/flat_map.hpp
@@ -1,0 +1,68 @@
+//                  ReactivePlusPlus library
+//
+//          Copyright Aleksey Loginov 2023 - present.
+// Distributed under the Boost Software License, Version 1.0.
+//    (See accompanying file LICENSE_1_0.txt or copy at
+//          https://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/victimsnino/ReactivePlusPlus
+//
+
+#pragma once
+
+#include <rpp/operators/map.hpp>
+#include <rpp/operators/merge.hpp>
+#include <rpp/defs.hpp>
+
+namespace rpp::operators::details
+{
+
+template<rpp::constraint::decayed_type Fn>
+struct flat_map_t 
+{
+    RPP_NO_UNIQUE_ADDRESS Fn m_fn;
+
+    template<rpp::constraint::observable TObservable>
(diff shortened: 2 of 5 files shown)


src/tests/rpp/test_flat_map.cpp
```diff
@@ -0,0 +1,126 @@
...
+                CHECK(mock.get_on_error_count() == 1);
+            }
+        }
+    }
+    /*
+    SECTION("flat_map doesn't produce extra copies")
+    {
+        copy_count_tracker verifier{};
+        auto obs = rpp::source::just<TestType>(std::move(verifier))
+                 | rpp::ops::flat_map([](copy_count_tracker&& verifier) { return rpp::source::just<TestType>(std::move(verifier)); });
```
CorentinBT: @victimsnino do you know why sources don't move objects through operators? Preventing me from taking verifier as an rvalue ref here

src/benchmarks/benchmarks.cpp
```diff
@@ -297,6 +297,22 @@ int main(int argc, char* argv[]) // NOLINT
                     | rxcpp::operators::subscribe<int>([](int v){ ankerl::nanobench::doNotOptimizeAway(v); });
             });
         }
+        SECTION("create+flat_map(just(v*2))+subscribe")
+        {
+            TEST_RPP([&]()
+            {
+                rpp::source::create<int>([](const auto& obs){ obs.on_next(1); })
+                    | rpp::operators::flat_map([](int v) { return rpp::source::just(v * 2); })
```
victimsnino: Could you use “create” there instead of just to measure there only “flat_map”, not “just” ? :)

src/rpp/rpp/operators/flat_map.hpp
```diff
@@ -0,0 +1,68 @@
...
+namespace rpp::operators::details
+{
+
+template<rpp::constraint::decayed_type Fn>
+struct flat_map_t 
+{
+    RPP_NO_UNIQUE_ADDRESS Fn m_fn;
+
+    template<rpp::constraint::observable TObservable>
+        requires std::invocable<Fn, rpp::utils::extract_observable_type_t<TObservable>>
```
victimsnino: There is also invocation result must be observable

src/rpp/rpp/operators/flat_map.hpp
```diff
@@ -0,0 +1,68 @@
...
+                | rpp::ops::merge();
+                
+    }
+
+    template<rpp::constraint::observable TObservable>
+        requires std::invocable<Fn, rpp::utils::extract_observable_type_t<TObservable>>
+    auto operator()(TObservable&& observable) &&
+    {
+        return std::forward<TObservable>(observable)
+                | rpp::ops::map(std::forward<Fn>(m_fn))
```
victimsnino: Should be move

src/rpp/rpp/operators/fwd.hpp
```diff
@@ -33,6 +33,10 @@ template<typename Fn>
     requires (!utils::is_not_template_callable<Fn> || !std::same_as<void, std::invoke_result_t<Fn, utils::convertible_to_any>>)
 auto map(Fn&& callable);
 
+template<typename Fn>
+    requires rpp::constraint::observable<std::invoke_result_t<Fn, utils::convertible_to_any>>
```
victimsnino: Add check over !utils::is_not_template_callable<Fn> as in map. It needed when passed function is purely template. For example `[](auto v) {return rpp::source::just(v); }` wouldn't satisfy your requirement

src/tests/rpp/test_flat_map.cpp
victimsnino: Because it would be moved IF possible, but no guarantees. If it would be only pure rvalue, then it means second repeated subscribe would see nothing :)

src/rpp/rpp/operators/flat_map.hpp
CorentinBT: It is already required [here](https://github.com/victimsnino/ReactivePlusPlus/pull/374/files#diff-a7a85d7aedddd82f15cd11012d2afd0e7fae88c55b3199b276f2500341bd1057R62) isn't it enough?

victimsnino: No, check this comment

https://github.com/victimsnino/ReactivePlusPlus/pull/374/files#r1225415597

src/tests/rpp/test_flat_map.cpp
CorentinBT: I see, I added some cases for copies then

src/benchmarks/benchmarks.cpp
CorentinBT: done

src/rpp/rpp/operators/flat_map.hpp
CorentinBT: done

src/rpp/rpp/operators/flat_map.hpp
CorentinBT: done

src/rpp/rpp/operators/fwd.hpp
CorentinBT: done, added a test too

src/tests/rpp/test_flat_map.cpp
victimsnino: if you would use create with std::move inside, then it would be moved inside lambda

src/rpp/rpp/operators/flat_map.hpp
```diff
@@ -22,8 +22,8 @@ struct flat_map_t
 {
     RPP_NO_UNIQUE_ADDRESS Fn m_fn;
 
-    template<rpp::constraint::observable TObservable>
-        requires std::invocable<Fn, rpp::utils::extract_observable_type_t<TObservable>>
+    template<rpp::constraint::observable TObservable, typename ValueType = rpp::utils::extract_observable_type_t<TObservable>>
```
victimsnino: I see your attempt with ValueType to avoid duplicates, but it would affect final binaries due to it would be compiled-in as template function with 2 params instead of one =C it is better to create special concept instead

src/tests/rpp/test_flat_map.cpp
```diff
@@ -105,22 +106,43 @@ TEMPLATE_TEST_CASE("flat_map", "", rpp::memory_model::use_stack, rpp::memory_mod
                 CHECK(mock.get_on_error_count() == 1);
             }
         }
+        SECTION("subscribe using flat_map with templated lambda")
```
victimsnino: i think, you can just place it in https://github.com/victimsnino/ReactivePlusPlus/pull/374/commits/a748a6f2997d75f16f5e5998d75f83696543da7d#diff-4ad61bca345fc2ef117d5617e164dca71340d801c636340f2c051a5bd4e6e9d2R37

src/tests/rpp/test_flat_map.cpp
```diff
@@ -105,22 +106,43 @@ TEMPLATE_TEST_CASE("flat_map", "", rpp::memory_model::use_stack, rpp::memory_mod
...
+TEST_CASE("flat_map copies/moves")
+{
     SECTION("flat_map doesn't produce extra copies")
     {
         copy_count_tracker verifier{};
-        auto obs = rpp::source::just<TestType>(std::move(verifier))
-                 | rpp::ops::flat_map([](copy_count_tracker&& verifier) { return rpp::source::just<TestType>(std::move(verifier)); });
-        SECTION("subscribe")
+        auto obs = rpp::source::create<copy_count_tracker>([verifier = std::move(verifier)](const auto& obs) { obs.on_next(verifier); })
+                 | rpp::ops::map([](copy_count_tracker verifier) { return std::move(verifier); }) // copy from source to map
```
victimsnino: try instead
```cpp
 auto obs = rpp::source::create<copy_count_tracker>([verifier = std::move(verifier)](const auto& obs) { obs.on_next(std::move(verifier)); })
```

CorentinBT: Though I subscribe two times here, so verifier will be in an invalid state in second subscription as it was already moved no?

src/rpp/rpp/operators/flat_map.hpp
CorentinBT: done

src/tests/rpp/test_flat_map.cpp
CorentinBT: done

src/tests/rpp/test_flat_map.cpp
CorentinBT: I changed to capture by ref and pass by value

victimsnino: > Though I subscribe two times here, so verifier will be in an invalid state in second subscription as it was already moved no?

No, verifier is special class that is valid after move =)

src/tests/rpp/test_flat_map.cpp
```diff
@@ -0,0 +1,139 @@
...
+                  })
+                | rpp::ops::subscribe(mock.get_observer());
+            SECTION("observer obtains values from underlying observables")
+            {
+                CHECK(mock.get_received_values() == std::vector{ 1, 3 });
+                CHECK(mock.get_on_completed_count() == 1);
+                CHECK(mock.get_on_error_count() == 0);
+            }
+        }
+        SECTION("subscribe using flat_map with error in middle")
```
victimsnino: Could you also add test over never in middle? to check that it is actually map + merge,  not map + concat =)

src/tests/rpp/test_flat_map.cpp
```diff
@@ -0,0 +1,139 @@
...
+            {
+                CHECK(mock.get_total_on_next_count() == 0);
+                CHECK(mock.get_on_completed_count() == 0);
+                CHECK(mock.get_on_error_count() == 1);
+            }
+        }
+    }
+}
+
+TEST_CASE("flat_map copies/moves")
```
victimsnino: I think, you can remove this test_case due to you are trying to test map + test merge at the same time, it is pretty hard. It is better to have corresponding tests to map and merge. I'm going to add them later.

src/tests/rpp/test_flat_map.cpp
CorentinBT: done

src/tests/rpp/test_flat_map.cpp
CorentinBT: yeah agree, done

victimsnino (dismissed):

src/rpp/rpp/operators/flat_map.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+template<rpp::constraint::decayed_type Fn>
+struct flat_map_t 
+{
+    RPP_NO_UNIQUE_ADDRESS Fn m_fn;
+
+    template<rpp::constraint::observable TObservable>
+        requires (std::invocable<Fn, rpp::utils::extract_observable_type_t<TObservable>> 
+                    && rpp::constraint::observable<std::invoke_result_t<Fn, rpp::utils::extract_observable_type_t<TObservable>>>)
+    auto operator()(TObservable&& observable) const &
```
victimsnino, line 28: Can you add test to cover this overloading? (it is enough to apply operator from variable in any test case)

CorentinBT: done

victimsnino (approved):

victimsnino:
Thanks for PR =)

PR #378, opened 2023-06-13
Add alias function to pipe operator

CorentinBT:


src/tests/rpp/test_observables.cpp
```diff
@@ -183,3 +183,39 @@ TEST_CASE("base observables")
...
+
+TEST_CASE("create observable works properly as observable")
+{
+    auto observable = rpp::source::create<int>([&](auto&& observer)
+    {
+        observer.on_next(1);
+        observer.on_completed();
+    });
+
+    SECTION("using const& variant")
```
victimsnino, line 189: actually they are both `const &` due to it is observable's ref qualifier. 
`const &` would be `observable.pipe(...)`
`&&` would be `std::move(observable).pipe()` or `rpp::source::create(...).pipe()`

CorentinBT: oh yes indeed I confused myself with operands, not sure why code coverage didn't decrease

victimsnino: Because it was totally "not created" due to it is templated function -> no instantiations -> no actual "creation"

src/tests/rpp/test_observables.cpp
```diff
@@ -183,3 +183,38 @@ TEST_CASE("base observables")
...
+        CHECK(pipe_operator_observer.get_total_on_next_count() == pipe_function_observer.get_total_on_next_count());
+        CHECK(pipe_operator_observer.get_on_error_count() == pipe_function_observer.get_on_error_count());
+        CHECK(pipe_operator_observer.get_on_completed_count() == pipe_function_observer.get_on_completed_count());
+    }
+    SECTION("using && variant")
+    {
+        mock_observer_strategy<int> pipe_operator_observer{};
+        mock_observer_strategy<int> pipe_function_observer{};
+
+        std::move(observable) | rpp::operators::subscribe(pipe_operator_observer.get_observer());
```
victimsnino: CI says that std::move is obsolete. Could you add // NOLINT or use rpp::source::create inlined?

CorentinBT: done

PR #411, opened 2023-08-26
Implement defer source

CorentinBT:


src/rpp/rpp/utils/function_traits.hpp
```diff
@@ -33,6 +33,11 @@ struct is_not_template_callable_t<R (*)(Args ...)> : std::true_type{};
 template<typename T>
 concept is_not_template_callable = is_not_template_callable_t<T>::value;
 
+template<typename F>
+concept is_no_argument_function = requires(F&& f) {
```
victimsnino: Is it just just `std::invocable<Fn>` ?

src/rpp/rpp/sources/defer.hpp
```diff
@@ -0,0 +1,50 @@
...
+//  Project home: https://github.com/victimsnino/ReactivePlusPlus
+
+#pragma once
+
+#include <rpp/sources/fwd.hpp>
+#include <rpp/observables/observable.hpp>
+
+namespace rpp::details
+{
+template<constraint::decayed_type Type, typename Factory>
```
victimsnino: Let's remove `Type` . It is deductible by factory

src/tests/rpp/test_defer.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include <rpp/sources/defer.hpp>
+#include <rpp/sources/from.hpp>
+#include <rpp/sources/empty.hpp>
+#include <rpp/sources/error.hpp>
+#include <rpp/sources/never.hpp>
+
+#include "mock_observer.hpp"
+
+TEST_CASE("defer on different sources")
```
victimsnino, line 22: Can we have test with mutable observable? For example, factory returns observable which emits incremented counter every subscribe. In case of direct usage such an observable we would see incremented counters while using defer - same initial value everytime

src/rpp/rpp/sources/defer.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace rpp::source
+{
+
+/**
+ * @brief Creates rpp::observable that calls the specified observable factory to create an observable for each new observer that subscribes.
+ *
+ * @tparam Factory the type of the observable factory
+ *
+ * @ingroup creational_operators
```
victimsnino, line 44: Can you add doxygen example?

src/rpp/rpp/sources/defer.hpp
CorentinBT: Hmm but this will invoke `std::invoke_result_t` two times and this might hurt compilation time a bit no?

src/rpp/rpp/sources/defer.hpp
CorentinBT: Any idea how to generate a marble equivalent to https://reactivex.io/documentation/operators/defer.html in rxmarbles? Looks complicated no?

src/tests/rpp/test_defer.cpp
CorentinBT: done

src/rpp/rpp/utils/function_traits.hpp
CorentinBT: done

src/rpp/rpp/sources/defer.hpp
victimsnino: Yeah, not sure if it is possible

src/rpp/rpp/sources/defer.hpp
victimsnino: Maybe, but i don't think so much. On other hand, it would not add template parameter and would not keep it everywhere while it is easy deducible, so, we would not affect resulting binary

src/rpp/rpp/sources/defer.hpp
victimsnino: I think, simple doxygen example like this
https://github.com/victimsnino/ReactivePlusPlus/blob/v2/src/examples/rpp/doxygen/just.cpp
with something similar to new added test and extended logging (inside factory function, indie observable and etc) would be enough

src/rpp/rpp/sources/defer.hpp
CorentinBT: done

src/rpp/rpp/sources/defer.hpp
CorentinBT: done

victimsnino (approved):

victimsnino:
Awesome!

PR #418, opened 2023-08-29
Interval source

CorentinBT:


src/tests/rpp/test_interval.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include <rpp/operators/as_blocking.hpp>
+#include <rpp/operators/subscribe.hpp>
+#include <rpp/operators/take.hpp>
+#include <rpp/sources/interval.hpp>
+#include <rpp/schedulers/current_thread.hpp>
+#include <rpp/schedulers/immediate.hpp>
+#include <rpp/schedulers/new_thread.hpp>
+
+#include "mock_observer.hpp"
+#include "test_scheduler.hpp"
```
victimsnino: Looks like you don't need it . And it is reason of compile fail

src/examples/rpp/doxygen/interval.cpp
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * \example defer.cpp
+ **/
+
+int main() // NOLINT
+{
+    //! [interval period]
+    rpp::source::interval(std::chrono::milliseconds(10), rpp::schedulers::immediate{})
+        | rpp::operators::take(3)
+        | rpp::operators::subscribe([](size_t v) { std::cout << v << "\n"; }, rpp::utils::rethrow_error_t{}, []() { std::cout << "On complete\n"; });
```
victimsnino: I think, in this case it is important WHEN v was obtained. Could you please add something like this?
https://github.com/victimsnino/ReactivePlusPlus/blob/v2/src/examples/rpp/doxygen/delay.cpp#L14C5-L14C53

src/examples/rpp/doxygen/interval.cpp
```diff
@@ -0,0 +1,32 @@
...
+    //         1 
+    //         2 
+    //         3
+    //         On complete
+    //! [defer from_iterable]
+    
+    //! [interval initial+period]
+    rpp::source::interval(std::chrono::milliseconds(5), std::chrono::milliseconds(10), rpp::schedulers::immediate{})
+        | rpp::operators::take(3)
+        | rpp::operators::subscribe([](size_t v) { std::cout << v << "\n"; }, rpp::utils::rethrow_error_t{}, []() { std::cout << "On complete\n"; });
```
victimsnino: in this case it is even much more important due to output is same

src/rpp/rpp/sources/interval.hpp
```diff
@@ -0,0 +1,97 @@
...
+#pragma once
+
+#include <rpp/sources/fwd.hpp>
+#include <rpp/observables/observable.hpp>
+
+namespace rpp::details
+{
+struct interval_schedulable
+{
+    rpp::schedulers::optional_duration operator()(auto&& observer, rpp::schedulers::duration period, size_t& counter) const
```
victimsnino, line 19: static?

victimsnino: i think, you can get observer like const auto&

src/rpp/rpp/sources/interval.hpp
```diff
@@ -0,0 +1,97 @@
...
+ * 
+ * @par Example:
+ * @snippet interval.cpp interval period
+ *
+ * @ingroup creational_operators
+ * @see https://reactivex.io/documentation/operators/interval.html
+ */
+template<schedulers::constraint::scheduler TScheduler>
+auto interval(rpp::schedulers::duration initial, rpp::schedulers::duration period, TScheduler&& scheduler) {
+    return interval_observable<TScheduler>{std::forward<TScheduler>(scheduler), initial, period};
```
victimsnino: std::decay_t<TScheduler>

src/rpp/rpp/sources/interval.hpp
```diff
@@ -0,0 +1,97 @@
...
+         return period;
+    }
+};
+
+template<typename TScheduler>
+struct interval_strategy
+{
+    using ValueType = size_t;
+
+    TScheduler scheduler;
```
victimsnino: RPP_NO_UNIQUE_ADDRESS

src/benchmarks/benchmarks.cpp
```diff
@@ -132,6 +132,30 @@ int main(int argc, char* argv[]) // NOLINT
                 rxcpp::observable<>::defer([] { return rxcpp::observable<>::iterate(std::array{123}, rxcpp::identity_immediate()); }).subscribe([](int v){ ankerl::nanobench::doNotOptimizeAway(v); });
             });
         }
+        SECTION("interval - interval + take(3) + subscribe + immediate")
+        {
+            TEST_RPP([&]()
+            {
+                rpp::source::interval(std::chrono::milliseconds(10), rpp::schedulers::immediate{}) | rpp::operators::take(3) | rpp::operators::subscribe([](size_t v) { ankerl::nanobench::doNotOptimizeAway(v); });
```
victimsnino: can we have less delay dependent benchmark? for example, interval with 1 nanosecond ?

src/examples/rpp/doxygen/interval.cpp
CorentinBT: done

src/examples/rpp/doxygen/interval.cpp
CorentinBT: done

src/rpp/rpp/sources/interval.hpp
CorentinBT: done

src/rpp/rpp/sources/interval.hpp
CorentinBT: done

src/rpp/rpp/sources/interval.hpp
CorentinBT: done

src/benchmarks/benchmarks.cpp
CorentinBT: done

victimsnino (approved):

PR #451, opened 2023-09-30
Reduce operator

CorentinBT:


src/rpp/rpp/operators/reduce.hpp
```diff
@@ -55,10 +55,11 @@ struct reduce_t : public operators::details::operator_observable_strategy<reduce
     using ResultValue = Seed;
 };
 
-template<rpp::constraint::decayed_type Seed, rpp::constraint::observer TObserver, rpp::constraint::decayed_type Accumulator>
+template<rpp::constraint::observer TObserver, rpp::constraint::decayed_type Accumulator>
 struct reduce_no_seed_observer_strategy
 {
     using DisposableStrategyToUseWithThis = rpp::details::none_disposable_strategy;
+    using Seed = std::invoke_result_t<Accumulator, utils::convertible_to_any, utils::convertible_to_any>;
```
CorentinBT: @victimsnino What do you think? Is it to remove Seed symbol from strategy template

victimsnino: At this point Seed is rpp::utils::extract_observer_type_t<TObserver>

victimsnino: But you proposal wouldn't work for templated lambdas

victimsnino: Same about `reduce_observer_strategy` and same about scan operator..

CorentinBT: Indeed, let me revert

victimsnino: Could you update to rpp::utils::extract_observer_type_t everywhere instead?

src/examples/rpp/doxygen/reduce.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * @example reduce.cpp
+ **/
+
+int main()
+{
+    //! [reduce]
+    rpp::source::just(1, 2, 3)
+        | rpp::operators::reduce(10, std::plus<int>{})
```
victimsnino: can we have reduce example with different types? like std::string

src/tests/rpp/test_reduce.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+#include <snitch/snitch.hpp>
+
+#include <rpp/operators/reduce.hpp>
+#include <rpp/sources/from.hpp>
+
+#include "copy_count_tracker.hpp"
+#include "mock_observer.hpp"
+
+TEMPLATE_TEST_CASE("scan scans values and store state", "", rpp::memory_model::use_stack, rpp::memory_model::use_shared)
```
victimsnino: reduce reduces? =)

src/rpp/rpp/operators/reduce.hpp
CorentinBT: done

src/examples/rpp/doxygen/reduce.cpp
CorentinBT: done

src/tests/rpp/test_reduce.cpp
CorentinBT: done

src/examples/rpp/doxygen/reduce.cpp
```diff
@@ -9,17 +12,17 @@
 int main()
 {
     //! [reduce]
-    rpp::source::just(1, 2, 3)
-        | rpp::operators::reduce(10, std::plus<int>{})
-        | rpp::operators::subscribe([](int v) { std::cout << v << std::endl; });
-    // Output: 16
+    rpp::source::just("b"s, "c"s)
+        | rpp::operators::reduce("a"s, std::plus<std::string>{})
```
victimsnino: No, my original idea was to make reduce working with different types. Like getting observalbe of ints and then reducing it to concatenated std::string =)

victimsnino: But NVM =)

victimsnino (dismissed):

PR #508, opened 2024-01-20
Add replay_subject

CorentinBT:


src/tests/rpp/test_subjects.cpp
```diff
@@ -255,4 +256,160 @@ TEST_CASE("serialized_subject handles race condition")
...
+}
+
+TEST_CASE("replay subject doesn't introduce additional copies")
+{
+    auto sub = rpp::subjects::replay_subject<copy_count_tracker>{};
+
+    SECTION("subscribe")
+    {
+        sub.get_observable().subscribe([](const copy_count_tracker& tracker){
+            CHECK(tracker.get_copy_count(), 1); // 1 copy to internal replay buffer
```
CorentinBT: @victimsnino any particular reason we enforce const lvalue ref on subjects strategies? In replay case, allowing forwarding ref would prevent this copy

src/rpp/rpp/subjects/replay_subject.hpp
```diff
@@ -0,0 +1,158 @@
...
+                state->time_points.push_back(now);
+            }
+
+            state->values.push_back(v);
+            state->state.on_next(v);
+        }
+
+        void on_error(const std::exception_ptr& err) const
+        {
+            std::unique_lock lock{state->mutex};
```
victimsnino: Why we should lock mutex there? mutex is guarding only lists, isn't it?

src/rpp/rpp/subjects/replay_subject.hpp
```diff
@@ -0,0 +1,158 @@
...
+                std::shared_ptr<subject_state<Type>>{m_state, &m_state->state}},
+            observer_strategy{m_state}};
+    }
+
+    template<rpp::constraint::observer_of_type<Type> TObs>
+    void on_subscribe(TObs&& observer) const
+    {
+        {
+            std::unique_lock lock{m_state->mutex};
+            for (const auto& value : m_state->values)
```
victimsnino, line 172: i think, we should do preprocessing from on_next there. Reason: duration. Let's imagine only one emisson happened 1 hour ago and we have duration for 5 minutes. If we are doing subscribe now, we shouldn't see any emissions

src/tests/rpp/test_subjects.cpp
victimsnino: This copy is result of copying to buffer. If you would forward value to buffer, it would be 1 move instead, isn't it?
Anyway, there is no enforcing of this thing, only requirement, that subject have to pass emisions as const lvalue to its observers due to we can't forward value to first one if we have multiple observers

src/rpp/rpp/subjects/replay_subject.hpp
```diff
@@ -0,0 +1,158 @@
...
+template<rpp::constraint::decayed_type Type>
+class replay_strategy
+{
+    struct replay_state
+    {
+        std::optional<size_t>                    count;
+        std::optional<rpp::schedulers::duration> duration;
+
+        std::list<Type>                                    values{};
+        std::list<rpp::schedulers::clock_type::time_point> time_points{};
```
victimsnino: i think, it is better to union it, else it would produce x2 "new" and x2 "delete" evertyime.

src/rpp/rpp/subjects/replay_subject.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+        std::mutex          mutex{};
+        subject_state<Type> state{};
```
victimsnino: please, inherit from it instead, my bad:
#509

src/tests/rpp/test_subjects.cpp
CorentinBT: Yes but I mean when a value comes inside a subject it can be forwarded no? I agree that emissions going out are mutlicasted to multiple observers hence the const lvalue ref.

src/rpp/rpp/subjects/replay_subject.hpp
CorentinBT: Depends do we want to make our replay_subject serialized too while we are at it?

src/tests/rpp/test_subjects.cpp
victimsnino: oh, yeah, sure, there is no such an requirement. just all previous subjects were simple multicasters

src/rpp/rpp/subjects/replay_subject.hpp
victimsnino: To be honest, i don't know. I think, we should have replay + serialized_replay subjects in this case?

src/tests/rpp/test_subjects.cpp
CorentinBT: done

src/rpp/rpp/subjects/replay_subject.hpp
CorentinBT: done, added both

src/rpp/rpp/subjects/replay_subject.hpp
CorentinBT: done

src/rpp/rpp/subjects/replay_subject.hpp
CorentinBT: done

src/rpp/rpp/subjects/replay_subject.hpp
CorentinBT: done

src/rpp/rpp/subjects/replay_subject.hpp
```diff
@@ -0,0 +1,222 @@
...
+            collect_and_on_next(std::move(v));
+        }
+
+        void on_next(const Type& v) const
+        {
+            collect_and_on_next(v);
+        }
+
+        void on_error(const std::exception_ptr& err) const
+            requires Serialized
```
victimsnino, line 119: tricky solution =)

src/rpp/rpp/subjects/replay_subject.hpp
```diff
@@ -0,0 +1,222 @@
...
+
+            std::unique_lock lock{state->serialized_mutex};
+            state->on_next(v);
+        }
+
+        template<typename T>
+        void collect_and_on_next(T&& v) const
+        {
+            state->collect(std::forward<T>(v));
+            state->on_next(v);
```
victimsnino: actually it's a bug. What if you've moved value during `collect`? You need to send .back() then instead

CorentinBT: done

PR #513, opened 2024-01-27
Timer source

CorentinBT:


victimsnino (approved):

victimsnino:
Thank you, @CorentinBT =)

src/rpp/rpp/sources/timer.hpp
```diff
@@ -0,0 +1,29 @@
...
+   }
+ *
+ * @param when time point when the value is emitted 
+ * @param scheduler the scheduler to use for scheduling the items
+ *
+ * @ingroup creational_operators
+ * @see https://reactivex.io/documentation/operators/timer.html
+ */
+template<schedulers::constraint::scheduler TScheduler>
+auto timer(rpp::schedulers::duration when, TScheduler&& scheduler)
```
victimsnino, line 25: BTW: @CorentinBT, isn't timer difference from interval that it accepts timepoint instead of duration??

PR #514, opened 2024-01-27
Time point timer

CorentinBT:


src/rpp/rpp/sources/timer.hpp
```diff
@@ -26,4 +26,19 @@ auto timer(rpp::schedulers::duration when, TScheduler&& scheduler)
...
+ * @param when time point when the value is emitted 
+ * @param scheduler the scheduler to use for scheduling the items
+ *
+ * @ingroup creational_operators
+ * @see https://reactivex.io/documentation/operators/timer.html
+ */
+template<schedulers::constraint::scheduler TScheduler>
+auto timer(rpp::schedulers::time_point when, TScheduler&& scheduler)
+{
+    return timer(when - rpp::schedulers::clock_type::now(), std::forward<TScheduler>(scheduler));
```
victimsnino: i think, it is incorrect solution. Example:
```cpp
auto obs = timer(now + 1day);
wait_for(20h);
obs.subscribe();
```

victimsnino: you should keep `when` inside observable and schedule to this timepoint at the moment of subscription

CorentinBT: done

victimsnino (approved):

PR #520, opened 2024-02-07
Distinct

CorentinBT:


src/tests/rpp/test_distinct.cpp
```diff
@@ -0,0 +1,71 @@
...
+    auto mock = mock_observer_strategy<int>{};
+    rpp::source::empty<int>() | rpp::operators::distinct() | rpp::ops::subscribe(mock);
+    CHECK(mock.get_received_values() == std::vector<int>{});
+    CHECK(mock.get_on_error_count() == 0);
+    CHECK(mock.get_on_completed_count() == 1);
+}
+
+TEST_CASE("distinct doesn't produce extra copies")
+{
+    copy_count_tracker::test_operator(rpp::ops::distinct(),
```
CorentinBT, line 57: @victimsnino do you know why this scenario doesn't compile? Compile error is a bit hard to follow, it seems that distinct strategy isn't constructible on this scenario

victimsnino: I think, copy_count_tracker is not hashable

CorentinBT: Ah yeah compile error was so hard to follow, moved the hashable constraint to a static_assert

victimsnino: Yeah, have to think about more understandable error messages =)

victimsnino (approved):

PR #521, opened 2024-02-10
Add tap operator

CorentinBT:


src/rpp/rpp/operators/tap.hpp
```diff
@@ -0,0 +1,158 @@
...
+ * @brief Register callbacks to inspect observable emissions and perform side-effects
+ *
+ * @param on_error error handler
+ *
+ * @ingroup utility_operators
+ * @see https://reactivex.io/documentation/operators/do.html
+ */
+template<typename OnNext                                       = rpp::utils::empty_function_any_t,
+         std::invocable<const std::exception_ptr&> OnError     = rpp::utils::empty_function_t<std::exception_ptr>,
+         std::invocable<>                          OnCompleted = rpp::utils::empty_function_t<>>
```
victimsnino: OnNext/OnCompleted should be omitted in this case and passed directly to details_tap_t

CorentinBT: done

PR #522, opened 2024-02-10
Improve error messages

victimsnino:


src/rpp/rpp/operators/distinct.hpp
```diff
@@ -49,10 +47,31 @@ struct distinct_observer_strategy
     bool is_disposed() const { return observer.is_disposed(); }
 };
 
+#define RESULT_TYPE_IF(res, ...) struct static_ensure { static_assert(__VA_ARGS__); using type = res; }; template<typename T> using result_value = typename static_ensure<T>::type
```
CorentinBT: if going with macro approach should we name it `ENABLE_OPERATOR_IF(...)`?

victimsnino: Good point, but it is also result type selector, not just “enabler”

victimsnino:
https://godbolt.org/z/PYv69hvh8

PR #523, opened 2024-02-11
Add zip operator

CorentinBT:


src/rpp/rpp/operators/zip.hpp
```diff
@@ -0,0 +1,129 @@
...
+    return details::zip_t<std::decay_t<TSelector>, std::decay_t<TObservable>, std::decay_t<TObservables>...>{
+        rpp::utils::tuple{std::forward<TObservable>(observable), std::forward<TObservables>(observables)...},
+        std::forward<TSelector>(selector)
+    };
+}
+
+/**
+ * @brief combines emissions from observables and emit tuple of items for each combination
+ *
+ * @marble zip_custom_selector
```
victimsnino: in this case it is just zip (this name should be unique to create unique images)

src/rpp/rpp/operators/zip.hpp
```diff
@@ -0,0 +1,129 @@
...
+public:
+    explicit zip_disposable(Observer&& observer, const TSelector& selector)
+        : combining_disposable<Observer, TSelector, Args...>(std::forward<Observer>(observer), selector)
+    {
+    }
+
+    auto& get_pendings() { return m_pendings; }
+
+private:
+    utils::tuple<std::list<Args>...> m_pendings{};
```
victimsnino: could we use dequeue instead? we don't need to insert/remove from middle

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -13,59 +13,32 @@
 #include <rpp/operators/fwd.hpp>
 
 #include <rpp/defs.hpp>
-#include <rpp/disposables/composite_disposable.hpp>
+#include <rpp/operators/details/combining_strategy.hpp>
 #include <rpp/operators/details/strategy.hpp>
-#include <rpp/operators/details/utils.hpp>
 #include <rpp/schedulers/current_thread.hpp>
```
victimsnino: not needed anymore

src/rpp/rpp/operators/zip.hpp
```diff
@@ -0,0 +1,129 @@
...
+//
+
+#pragma once
+
+#include <rpp/operators/fwd.hpp>
+
+#include <rpp/defs.hpp>
+#include <rpp/operators/details/combining_strategy.hpp>
+#include <rpp/operators/details/strategy.hpp>
+#include <rpp/schedulers/current_thread.hpp>
```
victimsnino: not needed anymore

src/rpp/rpp/operators/zip.hpp
```diff
@@ -0,0 +1,129 @@
...
+        const auto observer = disposable->get_observer_under_lock();
+        disposable->get_pendings().template get<I>().push_back(std::forward<T>(v));
+
+        disposable->get_pendings().apply([this, &observer](auto&... values) {
+            if ((!values.empty() && ...))
+            {
+                observer->on_next(disposable->get_selector()(std::move(values.front())...));
+                (..., values.pop_front());
+            }
+        });
```
victimsnino: This one can be extracted as separate function. Reason: reduce binary bloat: on_next function could be instantiated with const T&, T&, T&& and etc. For each of this one unique lambda would be instantiated

src/rpp/rpp/operators/details/combining_strategy.hpp
victimsnino, line 1: nice idea =)

src/rpp/rpp/operators/details/combining_strategy.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+    bool decrement_on_completed()
+    {
+        // just need atomicity, not guarding anything
+        return m_on_completed_needed.fetch_sub(1, std::memory_order::seq_cst) == 1;
+    }
+
+private:
+    value_with_mutex<Observer>      m_observer_with_mutex{};
+    RPP_NO_UNIQUE_ADDRESS TSelector m_selector;
```
victimsnino: Not sure if this one should be in base class. Actually it doesn't need it at all

src/rpp/rpp/operators/zip.hpp
CorentinBT: done

src/rpp/rpp/operators/zip.hpp
CorentinBT: done

src/rpp/rpp/operators/combine_latest.hpp
CorentinBT: done

src/rpp/rpp/operators/zip.hpp
CorentinBT: done

src/rpp/rpp/operators/zip.hpp
CorentinBT: Good point I didn't think about it, done

src/rpp/rpp/operators/details/combining_strategy.hpp
CorentinBT: done

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -13,59 +13,36 @@
...
-        // just need atomicity, not guarding anything
-        return m_on_completed_needed.fetch_sub(1, std::memory_order::seq_cst) == 1;
-    }
+    auto& get_values() { return m_values; }
 
 private:
-    value_with_mutex<Observer>                observer_with_mutex{};
-    rpp::utils::tuple<std::optional<Args>...> values{};
-    RPP_NO_UNIQUE_ADDRESS TSelector           selector;
+    RPP_NO_UNIQUE_ADDRESS TSelector m_selector;
```
victimsnino: Most probably crash on MSVC happens due to this line. Try to remove `RPP_NO_UNIQUE_ADDRESS` macro or place this member not first one

CorentinBT: good catch, done

src/tests/rpp/test_zip.cpp
```diff
@@ -0,0 +1,188 @@
...
+TEST_CASE("zip doesn't produce extra copies")
+{
+    SECTION("send value by copy")
+    {
+        copy_count_tracker verifier{};
+        auto               obs = verifier.get_observable()
+                 | rpp::ops::zip(
+                       [](copy_count_tracker&& verifier, auto&&) { return verifier; },
+                       rpp::source::just(1));
+        obs.subscribe([](copy_count_tracker) {});
```
victimsnino: NOLINT ?

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -13,59 +13,36 @@
...
 {
 template<rpp::constraint::observer Observer, typename TSelector, rpp::constraint::decayed_type... Args>
-class combine_latest_disposable final : public composite_disposable
+class combine_latest_disposable final : public combining_disposable<Observer, Args...>
 {
 public:
     explicit combine_latest_disposable(Observer&& observer, const TSelector& selector)
-        : observer_with_mutex{std::move(observer)}
-        , selector{selector}
+        : combining_disposable<Observer, Args...>(std::forward<Observer>(observer))
```
victimsnino: It should be move, not forward (same for zip)

src/rpp/rpp/operators/combine_latest.hpp
```diff
@@ -74,73 +51,21 @@ struct combine_latest_observer_strategy
...
-            disposable->get_observer_under_lock()->on_completed();
-        }
+        if ((vals.has_value() && ...))
+            observer->on_next(disposable->get_selector()(vals.value()...));
     }
 };
 
 template<typename TSelector, rpp::constraint::observable... TObservables>
-struct combine_latest_t
+struct combine_latest_t final : public combining_operator_t<combine_latest_disposable, combine_latest_observer_strategy, TSelector, TObservables...>
```
victimsnino: using?

CorentinBT: you mean to remove final?

src/rpp/rpp/operators/combine_latest.hpp
CorentinBT: done

src/rpp/rpp/operators/combine_latest.hpp
victimsnino: no, i mean we don't need struct in this case, just using to combining_operator_t...

PR #528, opened 2024-02-18
Add on_error_resume_next operator

CorentinBT:


src/rpp/rpp/operators/on_error_resume_next.hpp
```diff
@@ -0,0 +1,109 @@
...
+        bool is_disposed() const { return disposable.is_disposed(); }
+    };
+
+    template<rpp::constraint::decayed_type Selector>
+    struct on_error_resume_next_t : lift_operator<on_error_resume_next_t<Selector>, Selector>
+    {
+        template<rpp::constraint::decayed_type T>
+        struct operator_traits
+        {
+            using selector_observable_result_type =
```
victimsnino, line 67: i think it also should be static_assert(invocable<...>)

src/rpp/rpp/operators/on_error_resume_next.hpp
```diff
@@ -0,0 +1,109 @@
...
+                "Selector observable result type is not the same as T");
+
+            using result_type = T;
+
+            template<rpp::constraint::observer_of_type<result_type> TObserver>
+            using observer_strategy = on_error_resume_next_observer_strategy<TObserver, Selector>;
+        };
+
+        template<rpp::details::observables::constraint::disposable_strategy Prev>
+        using updated_disposable_strategy = rpp::details::observables::fixed_disposable_strategy_selector<1>;
```
victimsnino: actually it is not fixed<1>. What if Selector result would also set_upstream?

src/rpp/rpp/operators/on_error_resume_next.hpp
```diff
@@ -0,0 +1,109 @@
...
+     *
+     * @param selector callable taking a std::exception_ptr and returning observable to continue on
+     *
+     * @warning #include <rpp/operators/on_error_resume_next.hpp>
+     *
+     * @ingroup error_handling_operators
+     * @see https://reactivex.io/documentation/operators/catch.html
+     */
+    template<typename Selector>
+        requires (!utils::is_not_template_callable<Selector> || rpp::constraint::observable<std::invoke_result_t<Selector, std::exception_ptr>>)
```
victimsnino: in this case we don't need chaeck over `is_not_template` because we know that it should be callable with exception_ptr. This check needed when we don't know final type (for example, map)

src/rpp/rpp/operators/on_error_resume_next.hpp
victimsnino: To test it, you can return from selector observable with set_upstream in tests

CorentinBT: Or I can even test by putting nested on_error_resume_next, so I should rather use `atomic_dynamic_disposable_strategy_selector<1>` ?

src/rpp/rpp/operators/on_error_resume_next.hpp
CorentinBT: done

src/rpp/rpp/operators/on_error_resume_next.hpp
CorentinBT: done added a test with nested operators

PR #531, opened 2024-02-18
Finally operator

CorentinBT:


src/tests/rpp/test_finally.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+    auto mock = mock_observer_strategy<int>();
+    SECTION("observable with no emissions")
+    {
+        auto   obs     = rpp::source::create<int>([](const auto&) {
+        });
+        size_t invoked = 0;
+        SECTION("subscribe")
+        {
+            obs | rpp::operators::finally([&]() noexcept { ++invoked; })
```
CorentinBT, line 29: @victimsnino do you think we can relax the noexcept requirement?

victimsnino: Theoretically we can, but then we should handle it somehow properly. In my understanding it is not possible. Reason: dispose could happen from ANY place and we don't know where exception should be propogated. Also it could break internal disposable state....

src/rpp/rpp/operators/finally.hpp
```diff
@@ -0,0 +1,59 @@
...
+        template<rpp::details::observables::constraint::disposable_strategy Prev>
+        using updated_disposable_strategy = typename Prev::template add<1>;
+
+        RPP_NO_UNIQUE_ADDRESS LastFn last_fn;
+
+        template<rpp::constraint::decayed_type Type, rpp::constraint::observer Observer>
+        auto lift(Observer&& observer) const
+        {
+            observer.set_upstream(make_callback_disposable(last_fn));
+            return observer;
```
victimsnino: std::forward

src/tests/rpp/test_finally.cpp
CorentinBT: I see, I added a comment in doc to specify it

victimsnino (approved):

PR #537, opened 2024-02-27
timeout operator

victimsnino:


src/rpp/rpp/operators/timeout.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+            return rpp::observer<Type, timeout_observer_strategy<std::decay_t<Observer>, TFallbackObservable, container, TScheduler>>{std::move(ptr)};
+        }
+    };
+} // namespace rpp::operators::details
+
+namespace rpp::operators
+{
+    template<rpp::constraint::observable TFallbackObservable, rpp::schedulers::constraint::scheduler TScheduler /* = rpp::schedulers::immediate*/>
+    auto timeout(rpp::schedulers::duration period, TFallbackObservable&& fallback_observable, const TScheduler& scheduler /* = {}*/)
```
CorentinBT: Is it possible to add a variant with no fallback? And in case of timeout it just errors

victimsnino: Yeah, have plans to do it, just draft :)

PR #556, opened 2024-04-09
Forward selector exception in on_error_resume_next + expose test utils

CorentinBT:


src/rpp/rpp/operators/switch_on_next.hpp
```diff
@@ -62,8 +62,8 @@ namespace rpp::operators::details
 
         void on_error(const std::exception_ptr& err) const
         {
-            m_state->dispose();
             m_state->get_observer()->on_error(err);
+            m_state->dispose();
```
CorentinBT: I think dispose should happen after on_error/on_completed emissions, same thing for on_error_resume_next

victimsnino: Why? My point is that this state contains disposable for upstream, while observer is downstream. So, you firstly dispose upstream to correctly stop it and then forward value

CorentinBT: but then if one attaches a disposable callback he will first see the callback being called before the error emission?

victimsnino: If it comes from upstream - yes

victimsnino: But I don't know what is better in this case :)

victimsnino: OK, we can use your approach, but what about rest combining/merging operators in this case? They are using same approach, for example, merge

victimsnino: discussed in discord: let's do it as separate PR for ALL combining/merging operators

CorentinBT: done I reverted disposable changes for now until next PR

victimsnino (approved):

PR #558, opened 2024-04-10
Dispose before on_error/on_completed

CorentinBT:


src/rpp/rpp/sources/concat.hpp
```diff
@@ -90,6 +90,7 @@ namespace rpp::details
             if (state->itr.value() == std::cend(state->container))
             {
                 state->observer.on_completed();
+                state->dispose();
```
CorentinBT, line 93: not sure if this one is really needed, same thing if we need a dispose call on line 109

victimsnino: yeah, actually half of this calls would happen inside on_error/on_completed of downstream

src/tests/utils/disposable_observable.hpp
```diff
@@ -45,6 +45,38 @@ struct wrapped_observable_strategy_no_set_upstream
...
+        bool callback_called       = false;
+        auto observable_disposable = rpp::make_callback_disposable([&callback_called]() noexcept {
+            callback_called = true;
+        });
+        auto observable            = rpp::source::create<T>([&observable_disposable](auto&& obs) {
+            obs.set_upstream(observable_disposable);
+            obs.on_error(std::make_exception_ptr(std::runtime_error{""}));
+        });
+
+        op(observable) | rpp::ops::subscribe([](const auto&) {}, [&callback_called](const std::exception_ptr&) { CHECK(!callback_called); });
```
victimsnino, line 62: i think we also need to add check over CHECK(callback_called) after subscribe

PR #568, opened 2024-05-03
Add initial support for GRPC++

victimsnino:
Add client-side grpc support


src/extensions/rppgrpc/rppgrpc/client_reactor.hpp
```diff
@@ -0,0 +1,303 @@
...
+
+    public:
+        template<rpp::constraint::observable_of_type<Request> Observable, rpp::constraint::decayed_same_as<Observer> TObserver>
+        client_bidi_reactor(const Observable& messages, TObserver&& events)
+            : m_observer{std::forward<TObserver>(events)}
+            , m_disposable{messages.subscribe_with_disposable([this]<rpp::constraint::decayed_same_as<Request> T>(T&& message) {
+                std::lock_guard lock{m_write_mutex};
+                m_write.push_back(std::forward<T>(message));
+                if (m_write.size() == 1)
+                    Base::StartWrite(&m_write.front()); },
```
CorentinBT: Can zip operator be used to avoid using locks? One stream can emit write ready messages and other emit proto messages

src/extensions/rppgrpc/rppgrpc/utils/exceptions.hpp
```diff
@@ -0,0 +1,22 @@
...
+// Project home: https://github.com/victimsnino/ReactivePlusPlus
+//
+
+#pragma once
+
+#include <stdexcept>
+
+namespace rppgrpc::utils
+{
+    struct reactor_faield : public std::runtime_error
```
CorentinBT: typo?

src/extensions/rppgrpc/rppgrpc/client_reactor.hpp
```diff
@@ -0,0 +1,303 @@
...
+{
+    template<typename AsyncInMethod,
+             std::derived_from<AsyncInMethod> Async,
+             rpp::constraint::observable      Observable,
+             rpp::constraint::observer        Observer>
+    void add_client_reactor(member_bidi_function_ptr<AsyncInMethod, Observable, Observer> method,
+                            Async&                                                        async,
+                            grpc::ClientContext*                                          context,
+                            const Observable&                                             requests,
+                            Observer&&                                                    responses)
```
CorentinBT: Is it possible to add a `add_client_reactor` specialization to directly return a responses observable? Instead of client having to create a subject manually everytime?

src/extensions/rppgrpc/rppgrpc/server_reactor.hpp
```diff
@@ -0,0 +1,240 @@
...
+        template<rpp::constraint::decayed_same_as<Observer> TObserver>
+            requires (!rpp::constraint::decayed_same_as<TObserver, server_reader_reactor<Observer>>)
+        explicit server_reader_reactor(TObserver&& events)
+            : m_observer{std::forward<TObserver>(events)}
+        {
+            Base::StartSendInitialMetadata();
+            Base::StartRead(&m_read);
+        }
+
+        server_reader_reactor(server_reader_reactor&&) = delete;
```
CorentinBT: Why only move ctor is deleted?

src/tests/CMakeLists.txt
```diff
@@ -20,7 +20,12 @@ macro(add_test_target target_name module files)
   add_test_with_coverage(${TARGET})
 
   if (${module} STREQUAL rppqt)
-    rpp_add_qt_support_to_executable(${TARGET})
+  rpp_add_qt_support_to_executable(${TARGET})
```
CorentinBT: missing tab?

src/tests/utils/fakeit.hpp
```diff
@@ -0,0 +1,12162 @@
+#pragma once
```
CorentinBT: Is there a conan package for `fakeit`?

CorentinBT (commented):
Something to maybe consider is support for task cancellation. A grpc client context has a `TryCancel` method that could be called during early on_completed/on_error calls, allowing early termination on server side.

victimsnino:
> Something to maybe consider is support for task cancellation. A grpc client context has a `TryCancel` method that could be called during early on_completed/on_error calls, allowing early termination on server side.

@CorentinBT , Hmmmm, good idea, but loos like callback api doesn't have it. Only server could call Finish with not OK status

src/extensions/rppgrpc/rppgrpc/client_reactor.hpp
victimsnino: it is possible, but actually it means, that observable would be hot in this case and it can start to emit values till returning from such an function. This signature allows to avoid id. Also in this case user can select where to use serialized/replay/behavior subjects or not. In case of direct returning RPP should choose by itself

src/tests/utils/fakeit.hpp
victimsnino: Maybe, but I'm done some changes into it

src/extensions/rppgrpc/rppgrpc/client_reactor.hpp
victimsnino: Not sure if I get your point. Anyway, zip operator uses mutex itself :)

src/extensions/rppgrpc/rppgrpc/client_reactor.hpp
CorentinBT: Is it possible for a user to get a cold observable out of it instead? Subscription would then trigger the connection

victimsnino: Theoretically - yes, but what if user firstly want to emit values and then subscribe to observable later?

CorentinBT: Agree that is one possibility but maybe having both possibilities would help as it seems a bit tedious to always have to work with hot observables

victimsnino: Do you mean, any of this operations should start connection?

src/tests/utils/fakeit.hpp
victimsnino: added another mocking library

src/extensions/rppgrpc/rppgrpc/client_reactor.hpp
CorentinBT: It is already starting a connection immediately here no? Was wondering if the connection could be driven by suscription instead directly but of course it is always possible why your approach by wrapping it behind a `defer` I guess

src/tests/utils/fakeit.hpp
CorentinBT: Fyi gmock is also handy to mock grpc/protobuf they have an option to directly generate mocks when compiling protos

src/tests/rppgrpc/test_async_client.cpp
```diff
@@ -0,0 +1,211 @@
...
+        ClientCallbackReaderWriter stream_mock{};
+        trompeloeil::sequence      stream_sequence{};
+
+        REQUIRE_CALL(stub_mock, Bidirectional(trompeloeil::_, trompeloeil::_)).LR_SIDE_EFFECT({
+            reactor = _2;
+            CHECK(_1 == &ctx);
+            stream_mock.BindReactor(reactor);
+        });
+
+        const auto temp_reactor = new rppgrpc::client_bidi_reactor<Request, Response>();
```
victimsnino: @CorentinBT , what do you think about this way in this case? you just creating reactor, can manually subscribe observer/observable and then have to init it manually.

CorentinBT: Yes that's good indeed, why the `new` btw? We know for sure grpc manages it?

victimsnino: It is recommended way for grpc methods ¯\_(ツ)_/¯
Grpc is not going to manage it by its own, but looks lie we have to guarantee reactor is alive while it is not "done" in terms of grpc

PR #592, opened 2024-05-29
Fix disposing order

victimsnino:


CorentinBT:
should we use `test_operator_over_observable_finish_before_dispose` in operators tests?

PR #613, opened 2024-08-03
element_at operator

CorentinBT:
## Summary by CodeRabbit

- **New Features**
	- Introduced the `element_at` operator, allowing users to retrieve an item at a specified index from a sequence in ReactivePlusPlus.
	- Added a benchmarking section for `immediate_just` and `element_at` to enhance testing coverage.

- **Bug Fixes**
	- Improved error handling for the `element_at` operator, ensuring it manages edge cases and empty sequences correctly.

- **Tests**
	- Added a comprehensive suite of unit tests for the `element_at` operator, validating its functionality and performance across various scenarios.


src/rpp/rpp/operators/element_at.hpp
```diff
@@ -0,0 +1,99 @@
...
+        using updated_disposable_strategy = Prev;
+    };
+} // namespace rpp::operators::details
+
+namespace rpp::operators
+{
+    /**
+     * @brief Emit item located at specified `index` location in the sequence of items emitted by the source observable
+     *
+     * @marble take
```
victimsnino: Not take :)

src/tests/rpp/test_element_at.cpp
```diff
@@ -0,0 +1,110 @@
...
+                                              .send_by_copy = {.copy_count = 1, // 1 copy to final subscriber
+                                                               .move_count = 0},
+                                              .send_by_move = {.copy_count = 0,
+                                                               .move_count = 1} // 1 move to final subscriber
+                                          },
+                                          2);
+    }
+}
+
+TEST_CASE("take satisfies disposable contracts")
```
victimsnino: Not take

src/rpp/rpp/operators/element_at.hpp
```diff
@@ -0,0 +1,99 @@
...
+        RPP_NO_UNIQUE_ADDRESS TObserver observer;
+        size_t                          index{};
+        mutable size_t                  current{};
+
+        template<typename T>
+        void on_next(T&& v) const
+        {
+            if (current++ == index)
+            {
+                observer.on_next(std::forward<T>(v));
```
victimsnino, line 37: i think, we should call on_completed after on_next?

src/rpp/rpp/operators/element_at.hpp
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+
+        void on_error(const std::exception_ptr& err) const { observer.on_error(err); }
+
+        void on_completed() const
+        {
+            if (current <= index)
+            {
+                observer.on_error(std::make_exception_ptr(std::range_error{"index is out of bounds"}));
```
victimsnino: use custom rpp exceptions instead https://github.com/victimsnino/ReactivePlusPlus/blob/v2/src/rpp/rpp/utils/exceptions.hpp#L17

src/rpp/rpp/operators/element_at.hpp
```diff
@@ -0,0 +1,99 @@
...
+namespace rpp::operators::details
+{
+    template<rpp::constraint::observer TObserver>
+    struct element_at_observer_strategy
+    {
+        using preferred_disposable_strategy = rpp::details::observers::none_disposable_strategy;
+
+        RPP_NO_UNIQUE_ADDRESS TObserver observer;
+        size_t                          index{};
+        mutable size_t                  current{};
```
victimsnino, line 30: i think, it is possible to use 1 index like in take instead
 https://github.com/victimsnino/ReactivePlusPlus/blob/v2/src/rpp/rpp/operators/take.hpp#L28-L41

CorentinBT: I am not too sure for the on_error condition, just checking if `index == 0` doesn't work as it is decremented on each on_next. So a sequence of n values with `element_at(n)`  will never error

victimsnino: you can use index +1 as initial point, make -= 1 during on_next and emit value when zero was reached. In this case zero means "satisfied", but nvm, i can change it later

CorentinBT: Yeah though about this but makes logic a bit more complicated no?

src/rpp/rpp/operators/element_at.hpp
CorentinBT: done

src/tests/rpp/test_element_at.cpp
CorentinBT: done

src/rpp/rpp/operators/element_at.hpp
CorentinBT: indeed, done

src/rpp/rpp/operators/element_at.hpp
CorentinBT: nice, done

src/rpp/rpp/operators/element_at.hpp
victimsnino: yeah, it is why you can skip it

PR #616, opened 2024-08-18
retry_when operator

CorentinBT:
## Summary by CodeRabbit

- **New Features**
  - Introduced a `retry_when` operator for improved error handling and retry logic in reactive programming.
  - Added functionality to resubscribe to observables upon error notifications, enhancing user control over data flow.

- **Tests**
  - Added a comprehensive suite of unit tests for the `retry_when` operator to validate its functionality and ensure reliability in various scenarios.

- **Documentation**
  - Created an example demonstrating the use of the `retry_when` operator for conditional retries based on error types, serving as a practical guide for developers.


src/rpp/rpp/operators/retry_when.hpp
```diff
@@ -0,0 +1,158 @@
...
+        using preferred_disposable_strategy = rpp::details::observers::none_disposable_strategy;
+
+        RPP_NO_UNIQUE_ADDRESS mutable TObserver observer;
+        RPP_NO_UNIQUE_ADDRESS TObservable       observable;
+        RPP_NO_UNIQUE_ADDRESS Notifier          notifier;
+
+        template<typename T>
+        void on_next(T&&) const
+        {
+            observable.subscribe(std::move(observer));
```
CorentinBT: @victimsnino not too sure if we need any protection on making sure this on_next is called once? Maybe with a take(1) on notifier returned obs?

src/rpp/rpp/operators/retry_when.hpp
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        void on_error(const std::exception_ptr& err) const
+        {
+            observer.on_error(err);
+        }
+
+        void on_completed() const
+        {
+            observer.on_completed();
```
victimsnino: observer can be moved out in this place (for example, in case of just(1) it would be on_next + on_completed)

victimsnino: https://github.com/victimsnino/ReactivePlusPlus/blob/v2/src/rpp/rpp/operators/take_until.hpp#L62

src/rpp/rpp/operators/retry_when.hpp
victimsnino: yeah, i think it should be protected

src/rpp/rpp/operators/retry_when.hpp
```diff
@@ -0,0 +1,158 @@
...
+        struct operator_traits
+        {
+            using result_type = T;
+
+            template<rpp::constraint::observer_of_type<result_type> TObserver>
+            using observer_strategy = retry_when_impl_strategy<TObserver, TObservable, Notifier>;
+        };
+
+        template<rpp::details::observables::constraint::disposable_strategy Prev>
+        using updated_disposable_strategy = Prev;
```
victimsnino: actually not, it can be infinite amount of disposables in case of infinite amount of retries

victimsnino: let's make custom disposable instead like this (https://github.com/victimsnino/ReactivePlusPlus/blob/v2/src/rpp/rpp/operators/take_until.hpp#L38), pass it to original observer, accumulate all inner disposbles inside it and clear (https://github.com/victimsnino/ReactivePlusPlus/blob/b8b0ed589352bdb6bf05bb9b534ed0010b13cb40/src/rpp/rpp/sources/concat.hpp#L98) internals during moving from observable to callback and back. else it could grow internal container

CorentinBT: > actually not, it can be infinite amount of disposables in case of infinite amount of retries

Is there a way to amend the disposable test to cover this case?

victimsnino: i think it is possible. 
https://github.com/victimsnino/ReactivePlusPlus/blob/v2/src/tests/utils/disposable_observable.hpp#L104 there is tested fixed_disposable strategy, so, you have to test_operator_over_observable_with_disposable such an callable which would reproduce such an situation)

src/rpp/rpp/operators/retry_when.hpp
```diff
@@ -0,0 +1,194 @@
...
+#include <rpp/operators/details/strategy.hpp>
+#include <rpp/utils/constraints.hpp>
+#include <rpp/utils/utils.hpp>
+
+namespace rpp::operators::details
+{
+    template<rpp::constraint::observer     TObserver,
+             rpp::constraint::observable   TObservable,
+             rpp::constraint::decayed_type TNotifier>
+    struct retry_when_state : public rpp::composite_disposable
```
victimsnino: final

src/rpp/rpp/operators/retry_when.hpp
```diff
@@ -0,0 +1,194 @@
...
+    {
+        template<typename Observer, typename Observable, typename Notifier>
+        retry_when_state(Observer&& observer, Observable&& observable, Notifier&& notifier)
+            : observer(std::forward<Observer>(observer))
+            , observable(std::forward<Observable>(observable))
+            , notifier(std::forward<Notifier>(notifier))
+        {
+        }
+
+        RPP_NO_UNIQUE_ADDRESS TObserver   observer;
```
victimsnino, line 37: first member with RPP_NO_UNIQUE_ADDRESS for class with base class with vtable could cause UB on MSVC. place bool as first member or remove it for first member pls

src/rpp/rpp/operators/retry_when.hpp
```diff
@@ -0,0 +1,190 @@
...
+        void on_completed() const
+        {
+            locally_disposed = true;
+            state->observer.on_completed();
+        }
+
+        void set_upstream(const disposable_wrapper& d) { state->observer.set_upstream(d); }
+
+        bool is_disposed() const { return locally_disposed || state->is_disposed(); }
+    };
```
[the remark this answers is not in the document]
victimsnino: @CorentinBT , it is correct comment. Comparing to on_error in retry and on_completed in repeat, there on_next can be called multiple times, so, you also use locally_disposed inside on_next to avoid it

src/rpp/rpp/operators/retry_when.hpp
```diff
@@ -0,0 +1,190 @@
...
+{
+    /**
+     * @brief If an error occurs, invoke @p notifier and when returned observable emits a value
+     * resubscribe to the source observable. If the notifier throws or returns an error/empty
+     * observable, then error/completed emission is forwarded to original subscription.
+     *
+     * @param notifier callable taking a std::exception_ptr and returning observable notifying when to resubscribe
+     *
+     * @warning #include <rpp/operators/retry_when.hpp>
+     *
```
victimsnino, line 188: i think, doxygen example with usage would be pretty useful

src/rpp/rpp/operators/retry_when.hpp
```diff
@@ -0,0 +1,190 @@
...
+            , notifier(notifier)
+        {
+        }
+
+        RPP_NO_UNIQUE_ADDRESS TObserver   observer;
+        RPP_NO_UNIQUE_ADDRESS TObservable observable;
+        RPP_NO_UNIQUE_ADDRESS TNotifier   notifier;
+
+        std::atomic_bool is_inside_drain{};
+    };
```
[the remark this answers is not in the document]
victimsnino: fixed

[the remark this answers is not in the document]
victimsnino: @coderabbitai, could you check if it was fixed in code?

(checking how ai is good)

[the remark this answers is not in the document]
victimsnino: isnt observer second memeber now?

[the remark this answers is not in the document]
victimsnino: you are checking outdated version of code, check actual from the branch instead

src/rpp/rpp/operators/retry_when.hpp
CorentinBT: done added examples, for marble not sure its easily representable

src/examples/rpp/doxygen/retry_when.cpp
```diff
@@ -0,0 +1,63 @@
...
+            sub.on_next(std::string{"success"});
+            sub.on_completed();
+        }
+    })
+        | rpp::operators::retry_when([](const std::exception_ptr&) {
+              return rpp::source::timer(std::chrono::seconds{5}, rpp::schedulers::current_thread{});
+          })
+        | rpp::operators::subscribe([](const std::string& v) { std::cout << v << std::endl; });
+    // Source observable is resubscribed after 5 seconds on each error emission
+    //! [retry_when delay]
```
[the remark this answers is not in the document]
victimsnino: wow, looks like pretty valid point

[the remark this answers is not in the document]
CorentinBT: indeed!

src/examples/rpp/doxygen/retry_when.cpp
```diff
@@ -0,0 +1,63 @@
...
+ **/
+
+int main()
+{
+    //! [retry_when delay]
+    size_t retry_count = 0;
+    rpp::source::create<std::string>([&retry_count](const auto& sub) {
+        if (retry_count != 4)
+        {
+            sub.on_error({});
```
daixtrose, line 17: IIUC, the fundamental guarantees of observables are broken in this example:

After `on_error`, no other message must be emitted. This includes the emission of `on_completed`. See [the observable contract](https://reactivex.io/documentation/contract.html) for reference.

> **OnNext**
>    conveys an item that is emitted by the Observable to the observer
> **OnCompleted**
>    indicates that the Observable has completed successfully and that it will be emitting no further items
> **OnError**
>    indicates that the Observable has terminated with a specified error condition and that it will be emitting no further items 

I am having difficulty understanding how the semantics of `retry_when` fit into this contract. Operators usually act on the *output* of an observable. In this example, we have a restart functionality that *backpropagates* into the generator of a cold observable and acts on the observer itself. I am not happy with that design. 

Please explain why we need `retry_when`.

victimsnino: Actually not due to as required no any other messages emitted from **this** observable, but they would be emitted from new one (after resubscribe). Any observer obtained on_error wouldn't obtain any new messages. 
Actually we have chain like this
`observable->observer_inside_retry_when->final_observer`
when observable emits error, then only `observer_inside_retry_when` actually obtains error. Instead of forwarding error to `final_observer` it does attempt to subscribe **new** observer_inside_retry_when (let's say `observer_inside_retry_when_2`) observer to **new** observable_2. so old chain is partially destructed
`observable->observer_inside_retry_when->|     final_observer`
and it becomes like
`observable_2->observer_inside_retry_when_2->final_observer`

No any guarantees are broken in this case.

Why do we need it? For example,  to implement error handling loic with custom delaying. Like this

```c++
rpp::source::create<int>([](const auto& observer)
{
    // some hard job to construct state
     while(true) {
          observer.on_next(std::rand());
          if (get_current_cpu_temp() > 95) { // emulating some issues
              observer.on_error(.....);
              break;
          }
     }
    // some hard job to destruct state
})
| rpp::operators::retry_when([](const std::exception_ptr&) {
   return rpp::source::timer(std::chrono::seconds{5}, rpp::schedulers::new_thread{}); // kind of "ok, we are obtained error, it is ok, let's wait 5 seconds to make our cpu cooler and try again"
})
| rpp::operators::subscribe(...);
```

daixtrose: > no any other messages emitted from this observable, but they would be emitted from new one (after resubscribe)

That is precisely my problem with this operator. This only works and makes sense for cold observables. What happens if this operator is applied on a hot observable? It would subscribe to an observable that has been completed. Can we obtain a compile-time error to avoid this?

src/examples/rpp/doxygen/retry_when.cpp
```diff
@@ -0,0 +1,63 @@
...
+ * @example retry_when.cpp
+ **/
+
+int main()
+{
+    //! [retry_when delay]
+    size_t retry_count = 0;
+    rpp::source::create<std::string>([&retry_count](const auto& sub) {
+        if (retry_count != 4)
+        {
```
daixtrose, line 16: coderabbit.ai thinks there is a missing increment here, but I can no longer see its comments, so I hereby re-add the comment by hand.

src/examples/rpp/doxygen/retry_when.cpp
victimsnino: In this case hot observable should be in “error state”. For example, subject caches last error and emits it on new subscriptions. In this case it would be infinite loop, but it is up to user to control it.
Not sure if it is possible to add compile time error. Anyway it can be easily suppressed just via converting observable to dynamic version (and losing all meta info)

daixtrose: Given your arguments, I drop my request for a compile-time error message. Also, I accept your explanation:

The current design means that if the hot observable goes into an error state, you obtain zero emissions on consecutive subscriptions—unless you add a stateful relay like `subject`. This behavior follows the contract.  

I recommend that the difference in behavior for hot and cold observables is explicitly mentioned and explained by example in the documentation of `retry_when` to avoid people running into a pitfall.

victimsnino:
https://sonarcloud.io/code?id=victimsnino_ReactivePlusPlus&pullRequest=616&selected=victimsnino_ReactivePlusPlus%3Asrc%2Frpp%2Frpp%2Foperators%2Fretry_when.hpp&line=62
looks like your code (same as mine) is not testing non-recursive re-subscription

src/rpp/rpp/operators/retry_when.hpp
```diff
@@ -0,0 +1,197 @@
...
+                state->observer.on_error(std::current_exception());
+                return;
+            }
+        }
+    }
+
+    template<rpp::constraint::decayed_type TNotifier>
+    struct retry_when_t
+    {
+        TNotifier notifier;
```
victimsnino: NO_UNIQUE

src/examples/rpp/doxygen/retry_when.cpp
victimsnino: Yeah, sure, thank you!
@CorentinBT , could you please handle this documentation request?

src/rpp/rpp/operators/retry_when.hpp
CorentinBT: so its UB in msvc only if there are other members?

victimsnino: If inherited from base class with vtable. I think , it is partially overrides vtable in this case

victimsnino (approved):

Issue #617, opened 2024-08-19
puzzled : using observe_on(new_thread) & buffer in different order


0xMonad:
Thank you for creating and maintaining this project! It's very helpful to my work. I have encountered some problems in the process of using, and hope to get your help. Thank you again for your hard work!

### env: 
`gcc version 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04) C++20`

### code 
```cpp
#include <exception>
#include <fmt/core.h>
#include <fmt/format.h>
#include <rpp/rpp.hpp>

using namespace rpp;

int main()
{
    auto obs = source::just(1, 2, 3, 4, 5, 6, 7);

    // clang-format off
    obs    
        | ops::observe_on(schedulers::new_thread{})
        | ops::buffer(2)
        | ops::subscribe(
            [](const std::vector<int>& v) {
                fmt::println("A size: {} val: {}", v.size() ,fmt::join(v, "-"));
            }, 
            [](const std::exception_ptr& e) {}, 
            []() { fmt::println(" complete "); });

    obs | ops::buffer(2)
        | ops::observe_on(schedulers::new_thread{}, std::chrono::milliseconds(100))
        | ops::subscribe(
            [](const std::vector<int>& v) {
                fmt::println("B size {} size val: {}", v.size(), fmt::join(v, "-"));
            }, 
            [](const std::exception_ptr& e) {}, 
            []() { fmt::println(" complete "); });

    // clang-format on
    pause();

    return 0;
}
```

### output 
``` bash
A size: 2 val: 1-2
A size: 2 val: 3-4
A size: 2 val: 5-6
A size: 1 val: 7
 complete 
B size 2 size val: 1-2
B size 1 size val: 3
B size 1 size val: 4
B size 1 size val: 5
B size 1 size val: 6
B size 1 size val: 7
 complete 
```

### question
I think case A/B should have the same output. but actually different.

Or is there something wrong with my usage.

victimsnino:
Hey @GitJQ !
Thanks for using and you are welcome =)

yeah, this case looks pretty interesting... i would expect they should be exactly the same.. I've reproduced it locally, have to debug a bit

victimsnino:
UPD: looks like i've found where issue is appeared, going to prepare fix today

victimsnino:
Fixed in `v2` branch

PR #618, opened 2024-08-19
fix buffer capacity

victimsnino:
#617

## Summary by CodeRabbit

- **New Features**
  - Improved performance for the buffer operation by maintaining the capacity of the buffer after clearing.
  
- **Tests**
  - Added a new test case to validate the behavior of the buffer when accepting items by moving them, ensuring correct item processing without unnecessary copies.

### DIFF
--- a/src/rpp/rpp/operators/buffer.hpp
+++ b/src/rpp/rpp/operators/buffer.hpp
@@ -41,8 +41,11 @@ namespace rpp::operators::details
             m_bucket.push_back(std::forward<T>(v));
             if (m_bucket.size() == m_bucket.capacity())
             {
+                const auto capacity = m_bucket.capacity();
                 m_observer.on_next(std::move(m_bucket));
+
                 m_bucket.clear();
+                m_bucket.reserve(capacity);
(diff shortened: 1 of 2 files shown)


src/rpp/rpp/operators/buffer.hpp
```diff
@@ -41,8 +41,11 @@ namespace rpp::operators::details
             m_bucket.push_back(std::forward<T>(v));
             if (m_bucket.size() == m_bucket.capacity())
             {
+                const auto capacity = m_bucket.capacity();
                 m_observer.on_next(std::move(m_bucket));
+
                 m_bucket.clear();
+                m_bucket.reserve(capacity);
```
CorentinBT, line 48: Does it matter? I though clear doesn't relinquish memory unless you `shrink_to_fit`

victimsnino: i was expecting the same but in case of moving vector out it actually also moves-out capacity/size
https://godbolt.org/z/1fhhYW1xP

PR #628, opened 2024-08-30
fix infinite looping

victimsnino:
## Summary by CodeRabbit

- **New Features**
	- Enhanced the `is_disposed` method to provide a more comprehensive check by considering both stopped and disposed states.
	- Added a new test case for the `take_until` operator to prevent infinite loops in reactive streams.
	- Introduced new test cases for `concat`, `retry`, and `retry_when` operators to verify proper disposal behavior during looping scenarios.
	- Expanded test coverage for observable behavior regarding disposables on error and completion events.

- **Bug Fixes**
	- Improved the reliability of state checks, ensuring downstream logic interacts correctly with the updated method.
	- Refined state management in the `on_completed` method to enhance performance and clarity.

### DIFF
--- a/src/tests/rpp/test_take_until.cpp
+++ b/src/tests/rpp/test_take_until.cpp
@@ -13,12 +13,7 @@
 
 #include <rpp/observers/mock_observer.hpp>
 #include <rpp/operators/as_blocking.hpp>
-#include <rpp/operators/concat.hpp>
 #include <rpp/operators/finally.hpp>
-#include <rpp/operators/map.hpp>
-#include <rpp/operators/repeat.hpp>
-#include <rpp/operators/start_with.hpp>
-#include <rpp/operators/take.hpp>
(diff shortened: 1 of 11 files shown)


src/tests/rpp/test_take_until.cpp
```diff
@@ -239,3 +245,22 @@ TEST_CASE("take_until dispose after completion")
...
 }
+
+TEST_CASE("take_until infinite loop")
+{
+    rpp::source::create<int>([](auto&& obs) {
+        obs.on_next(1);
+        obs.on_completed();
+    })
+        | rpp::ops::map([](int) {
+              const auto delay = rpp::source::timer(std::chrono::seconds{1}, rpp::schedulers::current_thread{});
```
CorentinBT: maybe we can put 50ms as test will take 4s to execute otherwise

PR #685, opened 2024-11-21
Fix observer disposing after on_erorr/on_completed

victimsnino:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new `boolean_disposables_strategy` class for improved disposal state management.
	- Added a new `Boolean` mode to enhance observer control over disposal states.

- **Improvements**
	- Updated various observer strategies to use the new `Boolean` disposables mode, streamlining error handling and state management.
	- Removed redundant local disposal checks from several observer strategies, enhancing clarity and efficiency.

- **Bug Fixes**
	- Enhanced test cases to ensure proper validation of disposal states during observable operations.

### DIFF
--- a/src/rpp/rpp/observers/details/disposables_strategy.hpp
+++ b/src/rpp/rpp/observers/details/disposables_strategy.hpp
@@ -55,4 +55,17 @@ namespace rpp::details::observers
 
         static constexpr void dispose() {}
     };
+
+    class boolean_disposables_strategy
+    {
+    public:
+        static constexpr void add(const rpp::disposable_wrapper&) {}
+
+        bool is_disposed() const noexcept { return m_is_disposed; }
+
+        void dispose() const { m_is_disposed = true; }
+
+    private:
+        mutable bool m_is_disposed{};
+    };
 } // namespace rpp::details::observers
--- a/src/rpp/rpp/observers/details/fwd.hpp
+++ b/src/rpp/rpp/observers/details/fwd.hpp
@@ -25,7 +25,9 @@ namespace rpp::details::observers
         // No any disposables logic for observer expected
         None = 1,
         // Use external (passed to constructor) composite_disposable_wrapper as disposable
-        External = 2
+        External = 2,
+        // Observer just controls is_disposed or not but upstreams handled via observer_strategy
+        Boolean = 3
     };
 
     namespace constraint
@@ -47,6 +49,10 @@ namespace rpp::details::observers
      * @brief No any disposable logic at all. Used only inside proxy-forwarding operators where extra disposable logic not requires
      */
     struct none_disposables_strategy;
+    /**
+     * @brief Just control is_disposed or not via boolean and ignore upstreams at all
+     */
+    class boolean_disposables_strategy;
 
     /**
      * @brief Keep disposables inside dynamic_disposables_container container (based on std::vector)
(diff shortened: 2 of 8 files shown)


daixtrose:
I am sorry to have no solution but as an observer (sic!), I wonder why any RX C++ implementation ends in a mess of checking bools and (sub-)states like hell. Is this embedded in the problem domain? Is there no way to escape this? I mean, it is a paradox: a functional-style monad implementation is built out of state machines.

daixtrose:
You may remember me ranting about `dispose()` a while ago. Here, we have code that reads `return locally_disposed || disposable->is_disposed();`. I may resolve this in a decade, but this code looks wrong to me. I cannot express why.

victimsnino:
> I am sorry to have no solution but as an observer (sic!), I wonder why any RX C++ implementation ends in a mess of checking bools and (sub-)states like hell. Is this embedded in the problem domain? Is there no way to escape this? I mean, it is a paradox: a functional-style monad implementation is built out of state machines.

it is interesting point =D Yeah, anyway you have to check for some states, it is pretty natural - you are hiding state from user, but state should be handled somehow. Regarding this one: it appears just from optimization due to removing exact disposable state inside observer and changing it to lihtweight bool + external state. Anyway, yeah, i'm going to rewrite it a bit, just was doublechecking CI is ok

daixtrose:
> it is interesting point =D Yeah, anyway you have to check for some states, it is pretty natural - you are hiding state from user, but state should be handled somehow. Regarding this one: it appears just from optimization due to removing exact disposable state inside observer and changing it to lihtweight bool + external state. Anyway, yeah, i'm going to rewrite it a bit, just was doublechecking CI is ok

My problem is that the code does not hide the *payload* state that travels down the queue from the user; it is the extra janitor stuff that needs stateful expressions. This will get messy when we enter multi-threaded environments.

daixtrose:
> > I am sorry to have no solution but as an observer (sic!), I wonder why any RX C++ implementation ends in a mess of checking bools and (sub-)states like hell. Is this embedded in the problem domain? Is there no way to escape this? I mean, it is a paradox: a functional-style monad implementation is built out of state machines.
> 
> it is interesting point =D Yeah, anyway you have to check for some states, it is pretty natural - you are hiding state from user, but state should be handled somehow. Regarding this one: it appears just from optimization due to removing exact disposable state inside observer and changing it to lihtweight bool + external state. Anyway, yeah, i'm going to rewrite it a bit, just was doublechecking CI is ok

What happens if you try not to use bools but the lifetime of objects?

victimsnino:
> You may remember me ranting about `dispose()` a while ago. Here, we have code that reads `return locally_disposed || disposable->is_disposed();`. I may resolve this in a decade, but this code looks wrong to me. I cannot express why.

updated a bit to be a bit more clear, was just first version to be sure fix is valid

victimsnino:
> > it is interesting point =D Yeah, anyway you have to check for some states, it is pretty natural - you are hiding state from user, but state should be handled somehow. Regarding this one: it appears just from optimization due to removing exact disposable state inside observer and changing it to lihtweight bool + external state. Anyway, yeah, i'm going to rewrite it a bit, just was doublechecking CI is ok
> 
> My problem is that the code does not hide the _payload_ state that travels down the queue from the user; it is the extra janitor stuff that needs stateful expressions. This will get messy when we enter multi-threaded environments.

Not sure if i get your point

victimsnino:
> > > I am sorry to have no solution but as an observer (sic!), I wonder why any RX C++ implementation ends in a mess of checking bools and (sub-)states like hell. Is this embedded in the problem domain? Is there no way to escape this? I mean, it is a paradox: a functional-style monad implementation is built out of state machines.
> > 
> > 
> > it is interesting point =D Yeah, anyway you have to check for some states, it is pretty natural - you are hiding state from user, but state should be handled somehow. Regarding this one: it appears just from optimization due to removing exact disposable state inside observer and changing it to lihtweight bool + external state. Anyway, yeah, i'm going to rewrite it a bit, just was doublechecking CI is ok
> 
> What happens if you try not to use bools but the lifetime of objects?

It is correct question and it is why bools exists. Correct answer is "same". These bools are just optimization to NOT have unnecessary lifetime objects. But it is hidden from users

PR #732, opened 2026-01-08
Expanding parameter pack

FishAI:
MSVC with toolset v145 complaints about expanding this parameter pack

## Summary by CodeRabbit

* **Refactor**
  * Refined internal template constraints for operator handling to improve code consistency and maintainability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/rpp/rpp/operators/details/strategy.hpp
+++ b/src/rpp/rpp/operators/details/strategy.hpp
@@ -27,7 +27,7 @@ namespace rpp::operators::details
     class lift_operator
     {
     public:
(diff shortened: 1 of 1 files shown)


FishAI:
gcc does not seem to want it. Unfortunately, I have no gcc to test. Sorry for the failed build.

victimsnino:
Actually it can't be a fix. decayed_same_as accepts two arguments to compare. if you expand variadic inside this, than it would compare variadic with each one while current implementation provides list of types (==variadic) same as passed types

PR #757, opened 2026-08-16
share operator

btilds:
Adds `share()` as a shortcut for the `multicast<Subject>() | ref_count()` pipeline. It is templated over the subject with `publish_subject` as the default, mirroring `multicast()`, so `share<some_subject>()` works too.

Because it composes the existing `template_multicast_t` and `ref_count()`, the resulting observable type is identical to what `publish() | ref_count()` produces today, making it a drop-in replacement at existing call sites.

Covered by a new `share` test case in test_connectable_observable.cpp and a doxygen sample.

## Summary by CodeRabbit

* **New Features**
  * Added a `share` operator for multicasting observable values across multiple subscribers.
  * Automatically connects on the first subscription and disconnects after the last subscriber unsubscribes.
  * Supports configurable subject types, with publish behavior by default.
  * Added usage documentation and examples.

* **Tests**
  * Added coverage for shared subscriptions, lazy connections, subscriber lifetimes, and reconnection behavior.

### DIFF
--- a/src/examples/rpp/doxygen/share.cpp
+++ b/src/examples/rpp/doxygen/share.cpp
@@ -1,0 +1,35 @@
+#include <rpp/rpp.hpp>
+
+#include <iostream>
+
+/**
+ * @example share.cpp
+ **/
+int main() // NOLINT(bugprone-exception-escape)
+{
+    {
+        //! [share]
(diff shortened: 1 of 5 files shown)


AlexInLog (approved):
LGTM! thanks

